PR #23, opened 2025-05-23
column option for relevant model

larareichmann:


vdl_tools/shared_tools/climate_landscape/enrichment_pipeline.py
```diff
@@ -260,6 +260,7 @@ def run_relevance_model(
     model_name,
     idn,
     save_path,
+    column_text='text_for_relevance_model',
```
ztawilvdl, line 263: Should we make this an argument for the overall run_pipeline function?

larareichmann: but then how would it work for when you want to have more than 1 relevance model predicting on different text_cols?

ztawilvdl: It wouldn’t work with this implementation either 😀

larareichmann: just pushed that suggestion!

ztawilvdl: Did you commit it?

larareichmann: here... https://github.com/vibrant-data-labs/vdl-tools/pull/23/commits/4bbfbb0d9d7672a4fae95f9fa65f5dfa1b0da680

ztawilvdl: Oh  I see.

I meant [here](https://github.com/vibrant-data-labs/vdl-tools/pull/23/files#diff-393d801a5d3469d980da0f4b2d599811c979606f7f0d5eedc6f8fdace30ba5ceL416)

But I don't think it makes sense to do that now that I'm thinking about it.

ztawilvdl (approved):

PR #27, opened 2025-05-29
added option to use 'expanded text' in One Earth taxonomy mapping

ericberlow:
in the solutions sheet there is gpt expanded text that was not being used, so i added the option to use a different text field for the solutions.


ericberlow:
@larareichmann @foodwebster  - check if this is ok -  maybe there is a better general solution (e.g. we just rename the text field in the taxonomy that we want to use...) - or @foodwebster - was there a reason why you did the expanded solution text but then didn't use it in the taxonomy mapping?

PR #29, opened 2025-06-10
Lara/cb api exception

larareichmann:
@miket-dev I was having some errors while pulling CB and I got it to work with these fixes, just want to make sure that they make sense - thank you!

### DIFF
--- a/vdl_tools/scrape_enrich/crunchbase/api.py
+++ b/vdl_tools/scrape_enrich/crunchbase/api.py
@@ -90,8 +90,17 @@ def api_query_factory(url, default_fields, limit=None) -> Callable:
                 url,
                 json={"field_ids": fields, "query": filters, "limit": 1000, **next},
             )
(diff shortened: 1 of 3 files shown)


vdl_tools/scrape_enrich/crunchbase/api.py
```diff
@@ -90,8 +90,17 @@ def api_query(fields=default_fields, filters=list(), limit=limit):
...
                 json={"field_ids": fields, "query": filters, "limit": 1000, **next},
             )
+            if data.status_code != 200:
+                print(f"API error {data.status_code}: {data.text}")
+                break
+
+            try:
+                res_json = data.json()
+            except Exception as e:
+                print(f"Failed to decode JSON: {e}\nResponse text: {data.text}")
```
miket-dev, line 100: @larareichmann why this could happen in the successful response?

PR #33, opened 2025-08-05
Adding a more Strict Equity Justice list for "Underserved Communities"

ztawilvdl:


ztawilvdl:
Closed because I don't like the sensitivity of this and we no longer need it for OE Dashboards

ztawilvdl:
@ericberlow going to go back to this even though I didn't want to. So at least we can have a single source for how we are calculating it.

PR #35, opened 2025-08-08
Removing Unstructured Library from Webscraping

ztawilvdl:
Removing Unstructured Library from Webscraping. I think it will speed things up to cut down on CPU time and it makes installing and running `vdl_tools` on an ec2 instance simpler. (Opinion), I don't think it's as helpful for structuring the data now that the LLMs are getting so good.


vdl_tools/scrape_enrich/scraper/scrape_websites.py
```diff
@@ -545,23 +538,37 @@ def _format_scraped_sites(results):
...
-        'https://www.spicqyxdl.com/',
-        'https://elementalimpact.com/',
-        'https://elementalimpact.com/funding-opportunities/commercial-projects/'
-    ]
+    import json
+    import argparse
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--filename', type=str, default='')
+    parser.add_argument('--skip_existing', type=int, default=1)
```
miket-dev, line 546: better to use `BooleanOptionalAction` which will go to `True` or `False` based on the presence of the argument:
```
parser.add_argument('--skip_existing', action=argparse.BooleanOptionalAction, default=False)
```

vdl_tools/scrape_enrich/scraper/website_processor.py
```diff
@@ -154,17 +39,40 @@ def process_page_source(url: str, source: str):
...
-            logger.warn(f"URL {url} is empty")
-            return "empty"
-
-        return re.sub(pattern, " ", str(extract))
+        # Remove script and style elements
+        for script in soup(["script", "style"]):
+            script.decompose()
+
+            # Try to find main content areas
+            content_selectors = [
```
miket-dev, line 47: looks incomplete, these selectors assume that the website is created by someone who is familiar with semantic html tags or class names.
autogenerated css styles would be skipped (like css modules - `<div class="_mainContent__text_17a7f_163">`), and also sections/paragraphs which are put by default with various CMS. I could not confirm it on randomly picked websites, as it has a fallback to `body` though

miket-dev (approved):

vdl_tools/scrape_enrich/scraper/website_processor.py
ztawilvdl: Yeah, the fallback to `body` is what we had before right?

PR #37, opened 2025-08-25
Lara/subject kwds

larareichmann:
see usage in arpa-h repo here https://github.com/vibrant-data-labs/arpa_aitch/blob/535f9adad6fe05c71c276285bd49fde06ba22ff8/arpa_aitch/build_health_network.py#L103
(I tested build network with titles, keywords, and resolution with 1 or 2 levels)

It allows for keywords or "title" naming - either way it will store by default in "clusname" so that the player_attribute file doesn't need to be changed each time. 

@ztawilvdl do I need to set to True here to cash results? L 636
```
raw_kw = kw_cache.bulk_get_cache_or_run(
            given_ids_texts=ids_text_prompts,
            model=model,
            use_cached_result=False,
        )
```

### DIFF
--- a/vdl_tools/shared_tools/embed_texts_with_cache.py
+++ b/vdl_tools/shared_tools/embed_texts_with_cache.py
@@ -14,7 +14,7 @@ EMBEDDING_PROVIDER = {
 def embed_texts_with_cache(
     ids_texts: list[tuple[str, str]],
     use_cached_result: bool = True,
-    n_per_commit: int = 1000,
+    n_per_commit: int = 1500,
     max_workers=3,
     return_flat: bool = True,
     embedding_provider="openai",


vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
 
 
 
+def _suffix_from_col_name(colname: str) -> str:
+    m = re.search(r'(_L\d+)$', str(colname))
+    return m.group(1) if m else ""
```
ztawilvdl, line 434: Can you add a doc string for what this is trying to do? If there is a simpler solution than regex I'd prefer it. It can get slow at scale.

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
     textcol,
     clusattr='Cluster',
+    clusname=None,             # Use this for short sentence column name
     subject=None,
     model="gpt-4.1-mini",
+    n_tags = None,
+    n_entities=None,
+    sample_texts=None,
+    improve = False,
+    improve_model="o3-mini"  # Model for improving one-sentence summaries
```
ztawilvdl, line 501:
```suggestion
    improve_model="o3-mini"  # Model for improving one-sentence summaries
    centrality_col="ClusterCentrality"
```

This is an argument for the sampling function, so I'd make it an argument here to and make this the default

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
 ):
-
-    clus_centrality = "ClusterCentrality"
+    centrality_col = "ClusterCentrality"
```
ztawilvdl, line 503:
```suggestion
```

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -329,6 +330,68 @@ def add_text_below_token_limit(cdf, textcol, model, max_tokens=120000):
...
+    textcol,
+    clusattr="Cluster",
+    max_samples=50,
+    centrality_col="ClusterCentrality",
+):
+    """
+    Returns {cluster_id: [sampled_texts]}.
+
+    For each cluster:
+      - If size <= 2*max_samples: return the whole cluster sorted by centrality.
```
ztawilvdl, line 344: Let's say `size=75` so `75 <= (50*2)` so we'd choose `75`

But if `size = 110`, then we only select `min(50, 55)` == `50`.

That seems weird to me but let me know why I'm wrong.

vdl_tools/shared_tools/embedding_network/embedding_network.py
larareichmann: oh! I think I changed the code so that I didn't need this regex-

larareichmann: going to delete it

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -329,6 +330,68 @@ def add_text_below_token_limit(cdf, textcol, model, max_tokens=120000):
...
+        # Sample proportional to bin sizes
+        bin_counts = cdf["cent_bin"].value_counts(normalize=True)
+        n_per_bin = (bin_counts * n_samples).round().astype(int)
+
+        # Ensure the total allocation matches n_samples
+        diff = n_samples - n_per_bin.sum()
+        while diff != 0 and len(n_per_bin) > 0:
+            bin_to_adjust = n_per_bin.idxmax() if diff > 0 else n_per_bin.idxmin()
+            n_per_bin[bin_to_adjust] += 1 if diff > 0 else -1
+            diff = n_samples - n_per_bin.sum()
```
ztawilvdl, line 380: Why would this occur?

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
+        - Else falls back to: 'clus_sentence_short' / 'clus_sentence_long'
+        """
+    out = nodesdf.copy()
+    for col in (clusattr, short_col, long_col):
+        if col not in out.columns:
+            raise KeyError(f"Expected column '{col}' not found.")
+
+        # Minimal table for prompt
+    df_min = out[[clusattr, short_col, long_col]].drop_duplicates()
+    # print(df.shape)
```
ztawilvdl, line 460:
```suggestion
```

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
+            {"role": "user", "content": [{"type": "text", "text": review_prompt}]},
+        ],
+        response_format={"type": "json_object"},
+        temperature=1,
+        max_completion_tokens=10000,
+        top_p=1,
+        frequency_penalty=0,
+        presence_penalty=0,
+        store=False,
+    )
```
ztawilvdl, line 481: This is one where we could be caching.

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
+                          model="o3-mini",
+                          short_col=None,
+                          long_col=None):
+    """
+        Improves one-sentence summaries for a specific level and REPLACES the short column.
+        - If short_col/long_col are provided, uses them directly.
+        - Else if clusname is provided, infers columns as:
+             short = f"{clusname}{suffix_from(clusattr)}"
+             long  = short + "_long"
+        - Else falls back to: 'clus_sentence_short' / 'clus_sentence_long'
```
ztawilvdl, line 451: I don't think the second 2 points are happening anymore

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
-        sorted_cdf = cdf.sort_values("wtd_cc", ascending=False)
+    # Use provided sample_texts if given, else fallback to internal sampling
+    if sample_texts is not None:
+        clusters = sample_texts.keys()
+    else:
+        clusters = nodesdf[clusattr].unique()
+        sample_texts = {}
+        for clus in clusters:
+            cdf = nodesdf[nodesdf[clusattr] == clus].sort_values(centrality_col, ascending=False)
+            sample_texts[clus] = add_text_below_token_limit(cdf, textcol, model)
```
ztawilvdl, line 517: Make this a function that is like "`simple_sampling_for_texts`?

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -428,93 +558,69 @@ def get_cluster_sentences_from_text(
...
+            nodesdf, textcol, clusattr, max_samples=n_entities
+        )
 
-        ids_text_prompts.append((clus, get_text_for_entities(n_tags, cluster_texts)))
-        ids_text_sums.append((clus, get_text_for_summaries(cluster_texts, subject)))
+    # Iterate over the clusters and prepare the cluster texts
+    for clus, sampled_texts in cluster_samples.items():
+        # Use the sampled texts for each cluster
+        cluster_texts = sampled_texts  # Already sampled texts
+        ids_text_prompts.append((clus, format_kwds_prompt(n_tags, cluster_texts)))
```
ztawilvdl, line 623:
```suggestion
        ids_text_prompts.append((clus, format_kwds_prompt(n_tags, sampled_texts)))
```
Did I misunderstand why we need to set `cluster_text`?

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -584,17 +728,38 @@ def build_embedding_network(
...
+    naming_func = NAMING_FUNCTIONS.get(naming_strategy, get_cluster_kwdnames_from_text)
+    print('Using naming function:', naming_func.__name__)
+
+    sampled_texts_by_level = {}
+    # Assign cluster names with provided function or default
     if params.clusName is not None:
         for idx, clattr in enumerate(clusters):
             clName = params.clusName if idx == 0 else f"{params.clusName}_L{idx + 1}"
-            nodesdf = get_cluster_names_from_text(
+            print('sampling texts for', clattr, 'with n_entities', n_entities)
```
ztawilvdl, line 740: Can you replace all the print with `logger.info`

and import the logger from vdl_tools if needed?
`from vdl_tools.shared_tools.tools.logger import logger`

ztawilvdl (commented):
Looking good--added some comments. 

I'm a little confused by "short_col", "long_col" and the purpose for those.

vdl_tools/shared_tools/embedding_network/embedding_network.py
larareichmann: idk why I made it this complicated
what aobut "Take everything if the cluster is smaller than max_samples, else take exactly max_samples"

vdl_tools/shared_tools/embedding_network/embedding_network.py
larareichmann: this is to account for the rounding error when binning by centrality, to add or remove samples if needed due to the rounding in n_per_bin = (bin_counts * n_samples).round().astype(int)

larareichmann: I simplified it and rounded down instead, then distribute leftover samples with the bins more samples - avoids the while loop

vdl_tools/shared_tools/embedding_network/embedding_network.py
larareichmann: ok, changed to True

vdl_tools/shared_tools/embedding_network/embedding_network.py
```diff
@@ -366,41 +429,108 @@ def parse_keyword_response(response_obj):
...
+            },
+            {"role": "user", "content": [{"type": "text", "text": review_prompt}]},
+        ],
+        response_format={"type": "json_object"},
+        temperature=1,
+        max_completion_tokens=10000,
+        top_p=1,
+        frequency_penalty=0,
+        presence_penalty=0,
+        store=False,
```
larareichmann, line 480: store= False is something from openai that we are not utilizing

vdl_tools/shared_tools/embedding_network/embedding_network.py
ztawilvdl: Oh I see, but why not make this a caching object like the `PromptResponseCache` and call rather than using the OpenAI API?

larareichmann: this is a storing functionality from openai that I didn't fully understand- not our database caching

larareichmann:
> Looking good--added some comments.
> 
> I'm a little confused by "short_col", "long_col" and the purpose for those.

we have a prompt that attempts a title sentence, by seeing the cluster sampled texts, 
another makes a longer sentence - it's a short paragraph (not a lengthy summary)
we then send both (title+ context), of all clusters in the same openai call,  and ask it to rephrase the title sentence to get them to be more unique, without "loosing" the meaning of the cluster (that's why we provide the longer sentence)

makes sense?

vdl_tools/shared_tools/embedding_network/embedding_network.py
larareichmann: so I would need to create a class like  ImproveTitle(PromptResponseCacheSQL) with :
```
    def __init__(
        self,
        session,
        prompt_str,
        prompt_name=None,
        model="o3-mini",
    ):
        super().__init__(
            session=session,
            prompt_str=prompt_str,
            prompt_name=prompt_name,
        )
        self.model = model
```
  and pass the prompt to it , like that?

paths.ini
ztawilvdl, line 1: What's this for?

vdl_tools/shared_tools/keyword_extraction/subject_kwds_from_text.py
```diff
@@ -0,0 +1,345 @@
...
+        with open('./vdl_tools/shared_tools/keyword_extraction/keywords_extended.txt', 'r') as f:
+            new_kwds = [line.strip() for line in f if line.strip()]
+        logger.info(f"Read {len(new_kwds)} new keywords")
+        df_updated = append_new_kwords_to_dict(
+            kwd_dict_path="./vdl_tools/shared_tools/keyword_extraction/kwd_dic.csv",
+            list_new_kwds=new_kwds,
+            similarity_threshold=0.8,
+            emb_model="text-embedding-3-small",
+            output_csv_path="./vdl_tools/shared_tools/keyword_extraction/kwd_dic_expanded.csv"
+        )
```
ztawilvdl, line 345: I don't think we want to be saving the keywords in VDL tools since they should be saved in the project specific place. Right?

ztawilvdl (approved):
Small comments / questions

PR #40, opened 2025-09-15
added gov funder at the funding round level

ericberlow:
* also added detection of lists in the py2mapper attribute settings
* and fixed an `inplace` warning...

### DIFF
--- a/vdl_tools/tag2network/Network/ComputeClustering.py
+++ b/vdl_tools/tag2network/Network/ComputeClustering.py
@@ -70,7 +70,7 @@ def addLeidenClusters(nodesdf, nw, resolution=1.0, prefix='Cluster', min_clus_si
     nw : networkx.Graph, optional
         graph object, either linksdf or nw must be present. The default is None.
     resolution: float or list, optional
-        either a single value of clsutering resolution or a list of values, in which case clustering is hierarchical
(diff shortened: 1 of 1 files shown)


vdl_tools/scrape_enrich/prepare_crunchbase.py
```diff
@@ -218,6 +218,35 @@ def has_government_investor(investors_data):
...
+):
+    """ Add a boolean column to funding rounds dataframe indicating if any investor is a government investor """
+    investor_cols = [
+        'permalink',
+        'uuid',
+        'name',
+        'entity_def_id',
+        'investor_type',
+    ]
+    investor_df = orgs_investors_df[investor_cols]
```
ztawilvdl, line 233: This is unnecessary.  All you actually need is `investor_type` but I don't see a reason to create the new dataframe at all.

vdl_tools/scrape_enrich/prepare_crunchbase.py
```diff
@@ -218,6 +218,35 @@ def has_government_investor(investors_data):
             return True
     return False
 
+def add_gov_funder_to_rounds(
```
ztawilvdl, line 221: It doesn't actually add the column. It's more like `fr_round_has_gov_investor` but that is taken. 

That is doing the same thing.

PR #114, opened 2026-05-01
fixed text length, url mapping, addres parsing

larareichmann:
@ztawilvdl 
fix the mapping to the url when the org description is too short in "prepare for relevance model"
added a truncation to the text going to the relevance model to 500K tokens (doesn't hit often, but I was getting a few errors for very large context window in ed tracker)
fixed a parsing error in geotagging, I was getting a lot of error messages there and this seemed to have taken cared of it

### DIFF
--- a/vdl_tools/shared_tools/climate_landscape/enrichment_pipeline.py
+++ b/vdl_tools/shared_tools/climate_landscape/enrichment_pipeline.py
@@ -263,14 +263,13 @@ def prepare_for_relevance_model(df, max_workers=MAX_WORKERS, description_col='De
     # UPDATED: Uses the custom column if provided, defaults to 'Description'
     df['text_for_relevance_model'] = df[description_col]
 
-    df['Website'] = None
-
-    # df_missing_descriptions will have a Website either from Website_cb_cd or LinkedIn
     if df_missing_descriptions.shape[0] > 0:
-        df['Website'].fillna(df_missing_descriptions['Website'], inplace=True)
-    df['Website'].fillna(df['Website_cb_cd'], inplace=True)
-    df['Website Summary'] = df['Website'].map(summaries_for_missing_desc, None)
-    df['text_for_relevance_model'].fillna(df['Website Summary'], inplace=True)
+        df['Website Summary'] = df_missing_descriptions['Website Summary']
+        # For orgs with a short/missing description, prefer the website summary when available
+        has_website_summary = df['Website Summary'].notnull()
+        df.loc[has_website_summary, 'text_for_relevance_model'] = df.loc[has_website_summary, 'Website Summary']
+    else:
+        df['Website Summary'] = None
 
     # TO DO: Fix what happens when there is no website text
     # LinkedIn text? 990?
--- a/vdl_tools/shared_tools/model_caches/relevance_cache.py
+++ b/vdl_tools/shared_tools/model_caches/relevance_cache.py
@@ -4,11 +4,25 @@ import json
 from pathlib import Path
 
 import pandas as pd
+import tiktoken
 
 from vdl_tools.shared_tools.database_cache.database_utils import get_session
 from vdl_tools.shared_tools.openai.prompt_response_cache_sql import PromptResponseCacheSQL
 from vdl_tools.shared_tools.openai.openai_api_utils import get_completion
 from vdl_tools.shared_tools.tools.logger import logger
+
+MAX_TEXT_TOKENS = 500_000
+_ENCODING = "cl100k_base"
+
+
+def _truncate_text(text: str) -> str:
+    if not isinstance(text, str):
+        return text
+    encoding = tiktoken.get_encoding(_ENCODING)
+    tokens = encoding.encode(text)
+    if len(tokens) <= MAX_TEXT_TOKENS:
+        return text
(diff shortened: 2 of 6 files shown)


vdl_tools/shared_tools/model_caches/relevance_cache.py
```diff
@@ -4,12 +4,26 @@
...
+
+def _truncate_text(text: str) -> str:
+    if not isinstance(text, str):
+        return text
+    encoding = tiktoken.get_encoding(_ENCODING)
+    tokens = encoding.encode(text)
+    if len(tokens) <= MAX_TEXT_TOKENS:
+        return text
+    return encoding.decode(tokens[:MAX_TEXT_TOKENS])
+
```
ztawilvdl: I think we have a tool for this in openai_utils

Is `cl100k_base` the right encoding?

vdl_tools/shared_tools/climate_landscape/enrichment_pipeline.py
```diff
@@ -263,14 +263,13 @@ def prepare_for_relevance_model(df, max_workers=MAX_WORKERS, description_col='De
...
 
-    df['Website'] = None
-
-    # df_missing_descriptions will have a Website either from Website_cb_cd or LinkedIn
     if df_missing_descriptions.shape[0] > 0:
-        df['Website'].fillna(df_missing_descriptions['Website'], inplace=True)
-    df['Website'].fillna(df['Website_cb_cd'], inplace=True)
-    df['Website Summary'] = df['Website'].map(summaries_for_missing_desc, None)
-    df['text_for_relevance_model'].fillna(df['Website Summary'], inplace=True)
+        df['Website Summary'] = df_missing_descriptions['Website Summary']
```
ztawilvdl, line 267: How does this work if `df` is all organizations but `df_missing_descriptions` is only those that were missing a description. Aren't the shapes different?

vdl_tools/shared_tools/model_caches/relevance_cache.py
larareichmann: > cl100k_base
 is not but the difference maybe negligible ?

> I think we have a tool for this in openai_utils
let me check!

ztawilvdl (approved):

vdl_tools/shared_tools/model_caches/relevance_cache.py
larareichmann: Move truncation into get_completion so all API calls are protected.
Also, now the text allowance is calculated as context window minus prompt tokens and max
output tokens, all looked up from MODEL_DATA. Prompt tokens computed

vdl_tools/shared_tools/openai/openai_constants.py
```diff
@@ -1,4 +1,25 @@
...
+        "input_cost_per_token": 2.5 / 1_000_000,
+        "output_cost_per_token": 15.0 / 1_000_000,
+    },
+    "gpt-5.4-mini": {
+        "model_name": "gpt-5.4-mini",
+        "max_context_window": 400_000,
+        "max_output_tokens": 128_000,
+        "input_cost_per_token": 0.75 / 1_000_000,
+        "output_cost_per_token": 4.5 / 1_000_000,
+    },
```
ztawilvdl, line 22: Can you add `gpt-5.4-nano` now too please?

vdl_tools/shared_tools/openai/openai_api_utils.py
```diff
@@ -43,6 +43,37 @@ def get_num_tokens(text, model_name):
...
+    Returns (None, None) if the model is unknown.
+    """
+    if model_name in MODEL_DATA:
+        d = MODEL_DATA[model_name]
+        return d["max_context_window"], d["max_output_tokens"]
+    if model_name.startswith("ft:"):
+        base = model_name.split(":")[1]  # e.g. "gpt-4.1-mini-2025-04-14"
+        for key, d in MODEL_DATA.items():
+            if base.startswith(key):
+                return d["max_context_window"], d["max_output_tokens"]
```
ztawilvdl, line 74: We are missing the scenario where someone is using a model that isn't in `MODEL_DATA` which will return `None, None`. It should either fail OR have some conservative default like the `mins` from the model_datas.

Maybe like `400_000, 128_000`?

vdl_tools/shared_tools/climate_landscape/enrichment_pipeline.py
larareichmann: when I tested with data it seemed to work! 
I heard that " it's relying on pandas index alignment "
~~~~~
df_missing_descriptions is a subset of df with the same index values (e.g. rows [2, 5, 10]). When you do:

df['Website Summary'] = df_missing_descriptions['Website Summary']
Pandas aligns on index, so:

Rows 2, 5, 10 → get the website summary value
All other rows → get NaN
Then has_website_summary = df['Website Summary'].notnull() correctly selects only those rows to overwrite text_for_relevance_model.
~~~~~~

larareichmann (commented):
Thanks for the comments!

vdl_tools/shared_tools/openai/openai_api_utils.py
larareichmann: added this

PR #118, opened 2026-05-13
NZI: split early bucket into up_to_a / a_to_b, add equity-split + investor metrics, tighten survival gating

ericberlow:
## Summary

Three coordinated changes to the NZI funding-round processing pipeline that enable finer-grained stage-level analyses downstream (in `elemental-catalytic-capital`). They share a branch because the survival-classification cleanup (#2) depends on the bucket split (#1) being available.

### 1. Splitter refactor — `split_early_late_funding_rounds.py`

- **Bucket split**: the old `early` bucket (everything Pre-Seed → Series A) is split into two finer buckets so we can study the Seed → A scale gap separately from the A → B gap:
  - `up_to_a` — Pre-Seed + Seed rounds
  - `a_to_b`  — Series A + Early VC rounds
- `divide_funding_rows` now returns a **5-key dict** (`up_to_a`, `a_to_b`, `middle`, `late`, `exit`) instead of the previous 4-tuple.
- **New per-stage metrics** flattened into the per-company output (for `up_to_a` / `a_to_b` / `middle` only):
  - `{stage}_equity_raised`, `{stage}_nonequity_raised` — sum of equity vs. non-equity (`financing_type_nzi != "Equity"`) round amounts in the bucket. `equity + nonequity == amount_raised` when not trimming.
  - `{stage}_n_rounds_nonequity` — count of non-equity rounds in the bucket.
  - `{stage}_nonequity_types` — sorted list of unique `round_type_nzi` values among non-equity rounds (e.g. `["Grant", "Debt"]`).
- **New per-stage investor lists**: when `divided_funding_rows_and_flatten(..., processed_investor_df=...)` is passed, each pre-B bucket also gets `{stage}_investors` — a list of `{"id", "name", "primary_type"}` dicts covering every investor across the bucket's rounds, deduped. Defaults to `None` so existing callers are unaffected.
- **Boundary bug fix**: previously, `_find_first_index_at_or_above(stages, "a_to_b")` matched any elevated stage. For a company like `[Debt × 4, Late VC, Debt × 3, IPO]` this pulled the 4 leading Debt rounds into a synthesized `a_to_b` bucket — adding ~$1B per such company. Fixed by adding `_find_first_index_at_stage` and only creating the `a_to_b` bucket when there's an actual Series A / Early VC round.

### 2. Survival classification — `nzi_survival_rates.py`

- `NZI_SURVIVAL_STAGES` is now `["Seed", "Series A", "Series B"]` (was `["Seed", "Early VC", "Series B"]`).
- Adds an explicit **at-stage membership gate** in `_precompute_company_classifications`: a company is only eligible to be flagged `{stage}_succeeded` / `{stage}_failed` if it had a round of one of that stage's `at_stage_round_types`. Seed-only companies are no longer in the `series_a_*` cohort (previously the success-side check didn't enforce this, so the two flags were asymmetric).

### 3. Stage failure map — `nzi_zombie_companies_fail.py`

- Removes the redundant `"Grant"` entry from `STAGE_FAILURE_MAP` (it duplicated `Pre-Seed`).
- `"Pre-Seed"` is now a proper pre-Seed stage (graduates to Seed / Series A / Early VC).
- Adds `"Series B_Exit"` so we can measure graduation all the way to an exit event (IPO / SPAC / Acquisition / Buyout) rather than only to Series C.

### Tests

- `tests/test_nzi_funding_round_processing.py`: +143 lines covering the new bucket splits, equity/non-equity per-stage metrics, and per-stage investor lists.

## Why

The elemental-catalytic-capital project needed per-stage features richer than `amount_raised` and `num_rounds` to characterise the difference between companies that survive vs. fail at each transition. Specifically:

- Stage-specific equity vs. non-equity funding mix (grants/debt may sustain companies between equity rounds, but we couldn't measure that before).
- Stage-specific investor identity (which named funders back the survivors vs. the failures at each gate).

Both of these required splitting the early bucket and tightening cohort membership. The downstream consumer (`success_v_fail.py` in elemental-catalytic-capital) is already wired to use these new columns.

## Reviewer notes

- **API change**: `divide_funding_rows` returns a dict, not a tuple. Existing in-repo callers (`divided_funding_rows_and_flatten`) are updated.
- **New optional kwarg**: `divided_funding_rows_and_flatten(..., processed_investor_df=None)`. Default behaviour is unchanged — investor lists are only emitted if the kwarg is supplied.
- **Backwards compat for the 4-tuple legacy tests**: some pre-existing tests in `test_nzi_funding_round_processing.py` unpack a 4-tuple from `divide_funding_rows`; those tests are stale relative to the previous dict refactor and were not touched here. New tests use the dict / flatten API.
- **Performance**: investor-list emission iterates `round_investor_ids_nzi` per round. On the full NZI dataset (7k companies) the splitter completes in well under a minute end-to-end.

## Test plan

- [x] `pytest vdl_tools/scrape_enrich/netzero_insights/tests/test_nzi_funding_round_processing.py` — new tests pass.
- [x] End-to-end run of `elemental_catalytic_capital/process_data.py` produces the expected new columns on the processed companies file.
- [x] Spot-check of `equity_raised + nonequity_raised == amount_raised` identity (holds exactly when not trimming).
- [x] Spot-check on real data: known climatetech investors (Breakthrough Energy, DOE, NSF, Plug & Play, etc.) appear as top characteristic backers in the expected asset classes.


vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
```diff
@@ -223,22 +232,87 @@ def divide_funding_rows(
...
+        return None
+    lookup = {}
+    for _, row in processed_investor_df.iterrows():
+        inv_id = row.get("investor_id_nzi")
+        if pd.isna(inv_id):
+            continue
+        lookup[inv_id] = {
+            "id": inv_id,
+            "name": row.get("name_nzi"),
+            "primary_type": row.get("primary_type_nzi"),
```
ztawilvdl: Why only primary type?

vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
```diff
@@ -268,6 +362,9 @@ def divided_funding_rows_and_flatten(
         if x.startswith('has_') and x.endswith('_investor_calced_nzi')
     ]
 
+    investor_lookup = _build_investor_lookup(processed_investor_df)
+    include_investor_lists = investor_lookup is not None
+
```
ztawilvdl, line 395: I don't really understand why this is necessary

ztawilvdl (commented):
## Overall

Three coordinated changes look right in intent. Boundary-bug fix (`_find_first_index_at_stage` to distinguish "stage is exactly X" from "stage ≥ X") is the right shape, the at-stage gate added to `_precompute_company_classifications` correctly mirrors the check inside `did_company_fail`, and the new pre-B equity/non-equity + investor-list columns are cleanly gated by `INVESTOR_COUNT_BUCKETS` / `EQUITY_SPLIT_BUCKETS` / `INVESTOR_LIST_BUCKETS`.

New tests pass (2/2). Only internal callers within `process_nzi` consume `divide_funding_rows` / `divided_funding_rows_and_flatten`, so no external API churn.

A few things worth resolving before merge — none blocking but two would be useful as follow-ups.

## Correctness

### ⚠️ Edge case: leading non-equity rounds may now be silently dropped
The PR description correctly explains the old behavior (leading Debt rounds lumped into `a_to_b`, inflating it by ~$1B per affected company) and the fix excludes them — but in cases where there is **also** no `a_to_b` bucket, they're dropped entirely. Trace `[Debt × 4, Late VC, Debt × 3, IPO]`:

- `up_to_a = None` (no Pre-Seed/Seed in `[0,3]`)
- `first_a_to_b is None` → `a_to_b = None`
- `late = [Late VC, Debt × 3]`, `exit = [IPO]`
- → the 4 leading Debt rounds vanish from every bucket

This may be intentional (those companies are weird outliers and the old behavior was demonstrably wrong), but the comment at `split_early_late_funding_rounds.py:599-602` claims leading non-equity rows "get absorbed into `a_to_b`" — that's only true when `first_a_to_b is not None`. Suggest either documenting the drop explicitly or absorbing into `late` when `a_to_b` is absent. A test for `[Debt, Late VC]` (no Pre-Seed, no Series A) would lock the intended behavior.

### ⚠️ `Pre-Seed` cohort semantics silently changed
`nzi_zombie_companies_fail.py:49-52` tightens `at_stage_round_types` for `"Pre-Seed"` (drops `"Seed"`) and expands `graduation_stages` (adds `"Seed"`). Since `"Pre-Seed"` isn't in `NZI_SURVIVAL_STAGES`, the survival pipeline isn't affected — but any caller invoking `did_company_fail(..., at_stage="Pre-Seed")` directly will see different cohort membership and graduation criteria. Not called out in the PR summary; worth a sentence either in the PR body or in a code comment so future readers (and downstream consumers) know this is a deliberate semantic change.

### ⚠️ Pre-existing dead code in `divide_funding_rows`
Not introduced by this PR, but the early gates at `split_early_late_funding_rounds.py:211-212` and `:219-220` assign `split = [None, None, None, None, None]` and fall through — the assignment is immediately overwritten by `_split_on_first_late_round(...)` at line 225. These gates have been dead since the previous dict refactor. Either `return {...}` early (the apparent original intent) or drop the assignments entirely.

## Test coverage

- New tests are well-targeted: `test_flatten_equity_nonequity_split_per_stage` exercises the equity-identity invariant, multi-A rounds, no-non-equity buckets, and missing-bucket NaN behavior; `test_flatten_per_stage_investor_lists` exercises dedupe order, missing investor lookup, and the no-investor-df path.
- **25 of 27 pre-existing tests in `test_nzi_funding_round_processing.py` fail with `ValueError: too many values to unpack (expected 4)`.** Confirmed these also fail on `main` — they were broken by an earlier dict refactor, not this PR (PR body acknowledges this). However, the splitter is a ~700-line file with subtle boundary logic and effectively has 2 working tests. Since this PR makes substantive changes to `_split_on_first_late_round` and `_split_after_last_early_round`, it would be a good moment to repair them — each `early, middle, late, exit_rows = divide_funding_rows(...)` becomes `d = divide_funding_rows(...); early = d["up_to_a"]; ...` plus a per-test decision about which old `early` semantics map to `up_to_a` vs `a_to_b`. Even partial repair would be valuable.

## Minor

- `split_early_late_funding_rounds.py:443` — `_find_first_index_at_or_above` docstring still says `min_stage` is "one of `'early'`, `'middle'`, `'late'`, `'exit'`". Should now be `'up_to_a'` / `'a_to_b'` / `'middle'` / `'late'` / `'exit'`.
- `nzi_zombie_companies_fail.py:42-46` — example comment still uses `at_stage="Early VC"`. Could refresh to a current survival stage (`"Series A"` or `"Series B"`).
- `nzi_zombie_companies_fail.py:31` — `EXIT_TYPES` import lacks a trailing comma. Cosmetic.
- `_bucket_investors` includes unknown investor ids as `{"id": id, "name": None, "primary_type": None}`. Good defensive choice — if you want visibility into how often this happens on real data, log a count at the pipeline driver level.

## Performance / security
Splitter is per-company `groupby.apply`; investor list emission is O(rounds × investors_per_round). PR notes <1 min on 7k companies; fine. No new external input surfaces.

## Recommendation
Approve with two follow-ups, neither blocking:
1. Decide & document the `[Debt..., Late VC]` edge case — drop vs. absorb into `late`. Add a test either way.
2. Repair the 25 stale tests so the splitter has coverage backing its boundary logic before more changes land on top.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
```diff
@@ -183,32 +192,32 @@ def divide_funding_rows(
...
+        well as M&A events (Acquisition, Merger, Buyout).
 
     Raises
     ------
     ValueError
         If ``split_strategy`` is not a recognised value.
     """
     if not raised_equity_round(company_funding_rows):
-        split = [None, None, None, None]
+        split = [None, None, None, None, None]
```
ztawilvdl: Pre-existing dead code (not introduced by this PR, but visible here): this `split = [None, None, None, None, None]` is immediately overwritten by `_split_on_first_late_round(...)` / `_split_after_last_early_round(...)` at line ~225. Same on line 220. The early gates do nothing. Either `return {"up_to_a": None, ...}` here (the apparent original intent) or drop the assignments.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
```diff
@@ -430,35 +573,53 @@ def _split_on_first_late_round(company_funding_rows, stages):
...
+    late_start     = _find_first_index_at_or_above(stages, "late")
+    exit_start     = _find_first_index_at_or_above(stages, "exit")
+
+    # up_to_a window: rows before the first elevated row.
+    up_to_a_end = first_elevated - 1 if first_elevated is not None else last_idx
+
+    # Only emit up_to_a if there's at least one Pre-Seed/Seed round in
+    # the range. Otherwise, leading non-equity rows (debt, convertible) get
+    # absorbed into a_to_b — matching the old behavior where a convertible
+    # before Series A bundled into the same bucket as the A.
```
ztawilvdl: Comment is only accurate when `first_a_to_b is not None`. When the company has neither a `up_to_a` round nor an `a_to_b` round (e.g. `[Debt × 4, Late VC, Debt × 3, IPO]`), the leading Debt rows are absorbed into **nothing** — `up_to_a` is None (no Pre-Seed/Seed), `a_to_b` is None (no Series A), and the late bucket starts at the Late VC row, so the 4 leading Debt rounds vanish.

This may be intentional (those companies are weird outliers and the old behavior over-counted by ~$1B), but the current comment claims absorption that doesn't happen in this case. Either document the drop explicitly or absorb into `late` when `a_to_b` is absent. A test for `[Debt, Late VC]` (no Pre-Seed, no Series A) would lock the intended behavior.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/nzi_zombie_companies_fail.py
```diff
@@ -45,13 +46,9 @@
...
 STAGE_FAILURE_MAP = {
-    "Grant": {
-        "at_stage_round_types": ["Accelerator/incubator", "Grant", "Pre-Seed", "Seed"],
-        "graduation_stages": ["Series A", "Early VC"],
-    },
     "Pre-Seed": {
-        "at_stage_round_types": ["Accelerator/incubator", "Grant", "Pre-Seed", "Seed"],
-        "graduation_stages": ["Series A", "Early VC"],
+        "at_stage_round_types": ["Accelerator/incubator", "Grant", "Pre-Seed"],
+        "graduation_stages": ["Seed", "Series A", "Early VC"],
```
ztawilvdl, line 52: Cohort semantics for `Pre-Seed` change here: `at_stage_round_types` drops `"Seed"`, and `graduation_stages` adds `"Seed"`. `Pre-Seed` isn't in `NZI_SURVIVAL_STAGES` so the survival pipeline isn't affected, but any caller invoking `did_company_fail(..., at_stage="Pre-Seed")` directly will see different cohort membership and a different graduation bar. Worth a sentence in the PR body (or a code comment here) so it's not silent.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/nzi_zombie_companies_fail.py
```diff
@@ -28,6 +28,7 @@
     M_AND_A_SUCCESS_STAGE,
     TWO_YEARS_IN_DAYS,
     raised_equity_round,
+    EXIT_TYPES
```
ztawilvdl: Nit: missing trailing comma on the last import. Cosmetic, but existing import-block style in this file uses trailing commas.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/nzi_survival_rates.py
```diff
@@ -170,25 +174,42 @@ def _precompute_company_classifications(
...
+            # At-stage gate: a company can only be classified "succeeded" or
+            # "failed" at this stage if it actually had a round of one of the
+            # stage's `at_stage_round_types`. Otherwise both flags are False —
+            # this company isn't a member of this stage's cohort at all.
+            #
+            # `did_company_fail` already checks this internally (line 362 of
+            # nzi_zombie_companies_fail.py); we replicate the same gate here
+            # for the success side so the two flags are symmetric.
+            at_stage_round_types = set(mapping["at_stage_round_types"])
+            is_at_stage = bool(at_stage_round_types.intersection(company_round_types))
```
ztawilvdl: Correctness-wise this gate is right — it mirrors the check inside `did_company_fail` (zombie_companies_fail.py:362) and fixes the asymmetry where a Seed-only company that somehow graduated to Series B could be marked `series_a_succeeded=True` while `series_a_failed=False`.

Minor: `company_round_types` is computed once per company outside the stage loop (good), but `at_stage_round_types = set(mapping["at_stage_round_types"])` is rebuilt every stage every company. Pre-building `{stage: set(map["at_stage_round_types"])}` outside the company loop saves a few thousand set constructions on the full NZI dataset. Tiny — only worth it if you're already in here.

vdl_tools/scrape_enrich/netzero_insights/tests/test_nzi_funding_round_processing.py
```diff
@@ -1059,3 +1060,145 @@ def test_no_boundary_rounds_defaults_to_early_with_split_after_last():
     assert middle is None
     assert late is None
     assert exit_rows is None
```
ztawilvdl, line 1062: New tests are well-targeted. Two suggestions for additional coverage tied to splitter changes in this PR:

1. **Leading non-equity edge case** — `[Debt, Late VC]` or `[Debt × 4, Late VC, Debt × 3, IPO]`. Lock down whether those Debt rounds belong in `late` or are intentionally dropped (see comment on `split_early_late_funding_rounds.py:602`).
2. **Bucket-split semantics** — even one happy-path test confirming `up_to_a` / `a_to_b` boundaries for `[Pre-Seed, Seed, Series A, Series B]` would be useful.

Separately: the 25 pre-existing tests in this file fail with `ValueError: too many values to unpack (expected 4)` — confirmed on `main` too, so not introduced here, but the splitter effectively has ~2 working tests covering ~700 lines of subtle boundary logic. Each `early, middle, late, exit_rows = divide_funding_rows(...)` becomes `d = divide_funding_rows(...); early = d["up_to_a"]; ...` plus a per-test decision about whether the old `early` semantics map to `up_to_a` or `a_to_b`. Doesn't have to be this PR, but worth tackling before the next change on top of this file.

ztawilvdl (commented):
Inline comments anchoring the earlier summary to specific lines.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
ericberlow: good call. I didn't realize that elsewhere you had used primary and secondary list.

vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
ericberlow: Claude said it would be more efficient to just stop a list of investor id's at each stage and then look up other attributes.   I originally wanted it to add to the companies data frame a dictionary of investor id, name, and type(s).  Should we just do that to keep it simple?

ericberlow:
## PR Update — Reviewer Feedback Round

### `vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py`

**Bucket-boundary comment fix.** The comment at line ~599 (`# Only emit up_to_a if...`) overstated what happens to leading non-equity rows. Rewrote it to distinguish two cases:
- Leading Debt/Convertible rows are absorbed into `a_to_b` when a Series A exists.
- They are intentionally dropped when no Series A exists (e.g. `[Debt, Late VC, IPO]`) — documented why (avoids ~$1B over-count in the late bucket that the old code introduced).

**Dead-code fix.** The `split = [None]*5` assignments at the top of `divide_funding_rows` were immediately overwritten by `_split_on_first_late_round` / `_split_after_last_early_round`. Replaced both gates with explicit early returns of an all-`None` dict (the apparent original intent).

**Investor lookup — `primary_type` → `investor_types`.** The investor lookup previously emitted only `primary_type`; it now emits `investor_types`, a deduped list merging `primary_type_nzi` and `secondary_types_nzi` (primary first). Added a small `_collect_investor_types(primary, secondary)` helper. Unknown investors get `"investor_types": []` instead of `"primary_type": None`. Docstrings updated.

### `vdl_tools/scrape_enrich/netzero_insights/process_nzi/nzi_survival_rates.py`

**Loop hoist optimization.** `at_stage_round_types = set(mapping["at_stage_round_types"])` was being rebuilt on every (company × stage) iteration. Pre-built a `{stage: set(...)}` lookup once outside the company loop. Saves a few thousand set constructions on the full NZI dataset.

### `vdl_tools/scrape_enrich/netzero_insights/tests/test_nzi_funding_round_processing.py`

**Three new tests** appended (using the new dict-return API):

1. `test_leading_debt_dropped_when_no_pre_seed_and_no_series_a` — locks `[Debt, Late VC]`: leading Debt dropped.
2. `test_leading_debt_dropped_with_interleaved_debt_and_ipo` — locks `[Debt × 4, Late VC, Debt × 3, IPO]`: 4 leading Debts dropped; trailing 3 Debts absorbed into middle alongside Late VC; IPO in exit.
3. `test_happy_path_bucket_boundaries_pre_seed_through_series_b` — happy-path bucket boundaries for `[Pre-Seed, Seed, Series A, Series B]`.

**Existing test updated.** `test_flatten_per_stage_investor_lists` now exercises the new `investor_types` schema — fixture includes `secondary_types_nzi` covering: primary + secondary merge, secondary-only-missing, dedupe when secondary repeats primary, and unknown investors emitting `investor_types: []`.

*Note:* The 25 pre-existing tests that use the old 4-tuple unpacking still fail (they fail on `main` too — flagged by reviewer for a separate cleanup pass).

### `elemental_catalytic_capital/success_v_fail.py`

**Downstream consumer of the investor schema rename.**
- `_explode_investors_long` now reads `investor_types` (list) per investor instead of `primary_type` (string).
- `_aggregate_characteristic_investors` at `granularity="investor_type"` now explodes the `investor_types` list before grouping — so an investor classified as both Venture Capital (primary) and Corporate Venture Capital (secondary) contributes to **both** type-level aggregations. Cell baselines are computed before the explode, so unique-company baselines are unaffected by the double-counting an explode would otherwise introduce.
- Granularity option renamed: `"primary_type"` → `"investor_type"`. `run_all` callers updated.
- Output column renamed: `primary_type` → `investor_types` (at investor granularity, a list) / `investor_type` (at type granularity, a string after explode).

**Wilson metrics removed.**
- Deleted `_wilson_lower_bound` helper entirely.
- Dropped `success_rate_lower_bound_95` and `lift_lower_bound` columns from the output.
- Within-cell sort key switched from `success_rate_lower_bound_95` → `success_rate_smoothed`.
- Remaining metrics: `n_succeeded`, `n_failed`, `n_portfolio`, `n_companies_in_cell`, `baseline_rate`, `success_rate_raw`, `success_rate_smoothed`, `lift_smoothed`.
- Docstrings updated.

### `notebooks/characteristic_investors_reports.py`

Follow-on changes from the schema rename and Wilson removal:
- Top-investor tables now render `Smoothed rate` instead of `Wilson LB`.
- Type-pivot index and pivot column renamed `primary_type` → `investor_type`.
- Per-investor `Type` column renders the `investor_types` list (added `_fmt_types` helper for comma-joining).
- Section subtitles and report caveat text updated — Wilson-LB mentions replaced with smoothed-rate descriptions.

PR #127, opened 2026-05-29
Lara/cb query fixes

larareichmann:
One change to the cb_ query to allow orgs with no funding rounds, the other 2 changes were to solve bugs (1 time out/error during cb api query, the other was a column type error)


vdl_tools/scrape_enrich/crunchbase/api.py
```diff
@@ -99,10 +100,21 @@ def api_query(fields=default_fields, filters=list(), limit=limit):
...
+                    json={"field_ids": fields, "query": filters, "limit": 1000, **next},
+                )
+                if data.status_code == 200 and data.text.strip():
+                    break
+                wait = 2 ** attempt
+                print(f"\nCB API returned empty/invalid response (status {data.status_code}), retrying in {wait}s (attempt {attempt + 1}/{max_retries})...")
+                time.sleep(wait)
+            else:
+                print(f"\nCB API failed after {max_retries} retries, stopping pagination.")
+                break
```
ztawilvdl, line 117: What's this for?

vdl_tools/shared_tools/parquet_cache.py
```diff
@@ -186,7 +186,20 @@ def write_dataframe(
...
+        # that window slip through. Parse the offending column from the error,
+        # coerce it, and retry once.
+        import re
+        match = re.search(r"column (\w+) with type", str(e))
+        if match:
+            col = match.group(1)
+            logger.warning("Late-detected mixed-type column '%s', coercing to string and retrying.", col)
+            df = df.copy()
+            df[col] = df[col].map(_to_string)
+        table = pa.Table.from_pandas(df, preserve_index=False)
```
ztawilvdl: What's this for?

vdl_tools/scrape_enrich/crunchbase/api.py
larareichmann: when I was query CB it failed several times, even in EC2, so claude added a re-attempt to the API call - hasn't failed since (but tbh, I didn't see the print message that it was retrying).

vdl_tools/shared_tools/parquet_cache.py
larareichmann: when reading the combined parquet, I think there is a date column type (but has also another type) that was making giving a loading error and preventing from opening the file

larareichmann: apparently happens when trying to combine data from DB and API

larareichmann: ok, maybe this was not the right fix, since I got a log message when adding the cb permalink query:
"Late-detected mixed-type column 'created_at', coercing to string and retrying." 

So going back, the original bug happened when fetching fresh rows from the API and reading from the DB cache, the same column ended up with a mix of int and string representations of the same value.  This code here was patching instead of fixing.

The real fix:
Find where the int→string drift happens in the DB read, and normalize types there.

vdl_tools/scrape_enrich/crunchbase/api.py
ztawilvdl: can we replace print with `logger.warning`

vdl_tools/shared_tools/parquet_cache.py
ztawilvdl: Is DB read the postgres DB Read or Parquet cache?

ztawilvdl: @larareichmann can you try removing this hack + merging in this branch?
https://github.com/vibrant-data-labs/vdl-tools/pull/137
And testing?

ztawilvdl (approved):

vdl_tools/scrape_enrich/crunchbase/api.py
larareichmann: done!

PR #141, opened 2026-06-07
a_to_b bucket: split equity into first Series A vs post-first-A equity

ericberlow:
## What

In `split_early_late_funding_rounds.divided_funding_rows_and_flatten`, the **a_to_b** stage bucket now emits three additional columns that split `a_to_b_equity_raised` by *when* the equity was raised relative to the **first Series A / Early VC round** (the round that opens the bucket):

- **`a_to_b_first_a_raised`** — amount of that single first-A round. Series A and Early VC are interchangeable (`A_TO_B_TYPES`), so an Early-VC-only company's first Early VC counts as its first A. Taken directly (not summed), so an undisclosed amount stays `NaN`.
- **`a_to_b_post_first_a_equity_raised`** — sum of equity rounds *after* the first A (subsequent A bridges + any other equity before the first B), with the same undisclosed→NaN guard as `equity_raised`.
- **`a_to_b_n_rounds_post_first_a_equity`** — count of those. There is no `n_rounds_first_a` (always 1).

Normally `a_to_b_equity_raised == first_a_raised + post_first_a_equity_raised` and `a_to_b_n_rounds_equity == 1 + n_rounds_post_first_a_equity`.

## Why / scope
Lets the A→B survive-vs-fail model treat the headline Series A round separately from later bridges/equity. Only the `a_to_b` bucket is split; `up_to_a` equity is untouched.

## Verified
- Synthetic cases: Series A + bridge, Early-VC-only (first A = Early VC), undisclosed first A (→ NaN), single A (post = 0).
- 2,318 real a_to_b companies: the identity holds for **98.8%**. The ~1.2% exceptions have a pre-first-A equity round (an equity-tagged `Accelerator/incubator` round or `Equity crowdfunding` that lands in a_to_b when the company has no Pre-Seed/Seed boundary); that equity stays in `equity_raised` but in neither new column (recover via `equity − first_a − post`). ~0.1% have an a_to_b equity bucket with no Series-A-type round → the three columns are `None`.

Downstream (separate repo, elemental-catalytic-capital) wires these into the A→B model and drops the superseded `a_to_b_equity_raised` / `a_to_b_n_rounds_equity`.

### DIFF
--- a/vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
+++ b/vdl_tools/scrape_enrich/netzero_insights/process_nzi/split_early_late_funding_rounds.py
@@ -300,7 +300,14 @@ def divided_funding_rows_and_flatten(
         values among non-equity rounds). When the bucket exists,
         ``equity_raised + nonequity_raised == amount_raised`` (ignoring
         the NaN guard, which fires when every round of a given type in
-        the bucket was undisclosed). When
+        the bucket was undisclosed). The **a_to_b bucket only** additionally
+        gets ``a_to_b_first_a_raised`` (amount of the first Series A / Early VC
+        round), ``a_to_b_post_first_a_equity_raised`` (sum of equity rounds after
+        the first A — subsequent A bridges + any other equity before the first
+        B), and ``a_to_b_n_rounds_post_first_a_equity`` (their count). These split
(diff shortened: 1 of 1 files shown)


ericberlow:
since a to b equity raised was such a strong predictor of success - I split it into the actual first series a raised and then any post-first-a equity rounds.

PR #147, opened 2026-06-24
Migrate taxonomy_mapping engines to SQL prompt-response cache + per-project match_schema

ztawilvdl:
## Summary

Three connected changes to the `vdl_tools/shared_tools/taxonomy_mapping/` engines:

1. **Migrate every OpenAI call to the SQL prompt-response cache.** `hierarchical_taxonomy_mapping`, `check_taxonomy_coherence`, and `hierarchical_taxonomy_judge` now take a SQLAlchemy `session=` instead of an `openai.OpenAI` client, build their per-call work into `bulk_get_cache_or_run` batches against new `InstructorPRC` subclasses (`TaxonomyMatchCache`, `ScopeRecoveryCache`, `CoherenceAuditCache`, `JudgeCache`), and use Pydantic structured outputs via `responses.parse` instead of `chat.completions` + raw JSON mode. Re-running an unchanged pipeline is now a no-API run after the first successful pass.

2. **Per-project `match_schema` for the classifier.** `build_system_prompt` and `classify_entities` accept a Pydantic `match_schema` kwarg. When provided, it drives both the prompt's JSON shape description AND the API-layer strict-mode enforcement (via `TaxonomyMatchCache.response_model`), so the prompt prose and the structural constraint can't drift apart. Drivers define their own minimal `Match`/`MatchesResponse` classes:
   - **OE** uses `OneEarthMatch` (mode_of_operation as `Literal["direct", "enabling tech", "indirect"]`) and a confidence variant for the `confidence_threshold` knob.
   - **ED-tracker** uses `EdMatch` — index/evidence/reason only — because the ed taxonomy prompt has no modes and no confidence. This stops OpenAI strict-mode from forcing the model to emit `mode_of_operation`, which would spuriously trigger the indirect-fanout-stop and shorten walks. Default `MatchesResponse` is unchanged so any caller that doesn't opt in keeps working.

3. **Restore `temperature=0`.** The legacy `client.chat.completions.create` path set it explicitly; the new path was running at OpenAI's default (1.0) for non-reasoning models. `classify_entities` and `recover_unmatched` now take `temperature: float | None = 0`, threaded into `bulk_get_cache_or_run` via `**api_kwargs`. Automatically suppressed for reasoning models (`gpt-5*`) which reject the param.

Drop-in for OE / ED-tracker / Drawdown drivers; existing callers that don't pass `match_schema=` get the same `MatchesResponse` as before.

## What changed

| File | Change |
|---|---|
| `taxonomy_mapping/taxonomy_mapping_cache.py` | **New.** Pydantic schemas (`MatchesResponse`, `ScopeDecision`) + `InstructorPRC` subclasses (`TaxonomyMatchCache`, `ScopeRecoveryCache`). `TaxonomyMatchCache` accepts `response_model=` for per-project schemas. |
| `taxonomy_mapping/hierarchical_taxonomy_mapping.py` | Bulk-per-level rewrite of `classify_entities`. New helpers `_build_user_text`, `_build_given_id`, `_clean_matches`, `_seed_branch`. `recover_unmatched` migrated. `match_schema=` + `temperature=` kwargs threaded through. `_clean_matches` uses `getattr` so schemas without `mode_of_operation` / `confidence` work. Drops `build_openai_client`. |
| `taxonomy_mapping/check_taxonomy_coherence.py` | `AuditResponse` + `Gap` Pydantic schemas, `CoherenceAuditCache(InstructorPRC)`. `check_taxonomy_coherence` takes `session=` and uses a single bulk call. Drops `_audit_one` and the threadpool. |
| `taxonomy_mapping/hierarchical_taxonomy_judge.py` | `MatchScore`, `LevelMatches`, `AlignmentVerdict`, `JudgmentResponse` schemas + `JudgeCache(InstructorPRC)`. `run_judge_evaluation` takes `session=` and uses a single bulk call. Schema shape is now a stable list-of-LevelMatches (vs the old dynamic `<verdict_top_label>_alignment` key) so structured output works across configs. `JudgeConfig` drops `build_openai_client` + `verdict_key`. |
| `taxonomy_mapping/oe_hierarchical_taxonomy_mapping.py` | `map_to_oneearth` and `_walk_recovered_entities` take `session=` (drop `client=`). Adds `OneEarthMatch[WithConfidence]` schemas + `oneearth_match_schema()` factory; `build_oneearth_system_prompt` and `ONEEARTH_RESEARCH_SYSTEM_PROMPT` pass `match_schema=` based on `include_confidence`. Recovery scope prompt's JSON field renamed from `pillar` to `category` to match `ScopeDecision`. |
| `openai/prompt_response_cache_sql.py` | One-line touch (a default kwarg fix from a sibling PR). |

## Compatibility / migration

- `build_openai_client` is removed from `hierarchical_taxonomy_mapping` and `JudgeConfig`. Callers must use `with get_session() as session:` and pass `session=`. The ED-tracker and Drawdown drivers (in their own repos) have already been updated.
- Coherence engine signature: `check_taxonomy_coherence(client=...)` → `check_taxonomy_coherence(session=...)`. ED-tracker and Drawdown coherence-check drivers already updated.
- Judge JSON output key changed from `<pillar>_alignment` to `alignment_verdict` (and `matches` is now a list of `{level, matches: [...]}` instead of a dict keyed by level). Downstream DataFrame columns are unchanged, so reports and aggregates carry through transparently. Existing cached judge responses from before this PR are invalidated by the new `prompt_id` (new system prompt + new Pydantic schema).

## Quality check

Judged a 464-uid subset (intersection of a new 500-entity sample + the 2026_04_17 ed_tracker baseline) with the same `JudgeConfig`:

- **98.0% of shared (entity, level, match) score points unchanged**.
- **94.8% of verdicts unchanged**.
- Net verdict change: **+9 `correct`** (mostly from previously-refused entities the new pipeline now classifies; 11 of the 12 verdict scope-flips are `no_pillar_expected → correct`).
- 7-entity verdict regression and 15 small score regressions (vs 14 improvements) on 1,451 shared keys — within noise.
- Pillar-level mean score: 0.9753 → 0.9599 (small dip; trade-off for the increased recall).

## Test plan

- [ ] Run a small slice (~50 entities) through OE `map_to_oneearth` via a fresh session and confirm the per-row frame shape matches the legacy schema (`Pillar / Sub-Pillar / Solution / Sub-Term` + `deepest_match / leaf_definition / mode_of_operation / evidence / reason / confidence`).
- [ ] Run the same slice with `read_from_cache=False` and confirm fresh API calls populate the cache; re-run and confirm 100% cache hit (zero API calls).
- [ ] Run ED-tracker driver with `match_schema=EdMatchesResponse`; confirm `mode_of_operation` column in the per-row output is empty for every row (schema doesn't include the field → model can't emit it).
- [ ] Run the coherence checker on a small taxonomy and confirm the audit DataFrame matches the legacy column shape.
- [ ] Run the judge on a small per-row file and confirm `_quality_scored.xlsx`, `_quality_<top>_verdicts.xlsx`, `_quality_summary.xlsx`, and `_quality_report.md` are written.
- [ ] Confirm reasoning models (`gpt-5*`) work without errors — the `temperature` kwarg should be suppressed automatically.

### DIFF
--- a/vdl_tools/shared_tools/openai/openai_api_utils.py
+++ b/vdl_tools/shared_tools/openai/openai_api_utils.py
@@ -1,3 +1,4 @@
+import contextlib
 import logging
 
 from openai import AsyncOpenAI, OpenAI
@@ -178,6 +179,107 @@ def get_completion(
         return response
     return response.output_parsed
 
+
+@contextlib.contextmanager
+def spy_openai(send: bool = True, printer=print, show_response: bool = True):
+    """Temporarily print every ``CLIENT.responses.parse`` call's payload.
+
+    Wrap any code that ends up calling the Responses API — ``get_completion``,
+    or the taxonomy-mapping / judge / coherence caches (which all funnel
+    through it) — and this prints the full request each call builds: the
+    ``model``, every extra kwarg (``temperature``, ``reasoning``, ...), the
+    ``text_format`` response class, and the ``input`` messages (or the bare
+    user string + ``instructions`` for reasoning models). The patch is
+    restored on exit.
+
+    Parameters
+    ----------
+    send : bool
+        ``True`` (default): print the request, call the real API, and
+        (if ``show_response``) print ``response.output_parsed``.
+        ``False``: print the request and raise *before* sending — no API
+        call. Use this with a single call; inside a multi-call loop the
+        raise is caught per call (recorded as an error) and the loop
+        continues, so you'd see every request but get no results.
+    printer : callable
+        Where to send output. Defaults to ``print``; pass e.g.
+        ``logger.info`` to route elsewhere.
+    show_response : bool
+        When ``send=True``, also print the parsed response. Default True.
+
+    Examples
+    --------
+    >>> from vdl_tools.shared_tools.openai.openai_api_utils import (
+    ...     get_completion, spy_openai,
+    ... )
+    >>> with spy_openai():
+    ...     get_completion(prompt="...", model="gpt-4.1", text="...",
+    ...                    text_format=MySchema)
+
+    >>> # see the request without spending a call:
+    >>> with spy_openai(send=False):
+    ...     try:
+    ...         cache.get_cache_or_run(given_id="x", text="...",
+    ...                                read_from_cache=False, write_to_cache=False)
+    ...     except RuntimeError:
+    ...         pass
+    """
+    orig = CLIENT.responses.parse
+    # ``parse`` is normally a class method (not an instance attribute), so
+    # restoring means deleting the instance-attribute shadow we add below
+    # — not reassigning, which would leave a stale bound method behind.
+    had_own_parse = "parse" in vars(CLIENT.responses)
+
+    def _wrapper(*args, **kwargs):
+        lines = ["", "=" * 72, "OpenAI responses.parse — request", "=" * 72]
+        lines.append(f"model       : {kwargs.get('model')}")
+        for k, v in kwargs.items():
+            if k in ("model", "input", "instructions", "text_format"):
+                continue
+            lines.append(f"{k:<12}: {v!r}")
+        tf = kwargs.get("text_format")
+        lines.append(f"text_format : {getattr(tf, '__name__', tf)}")
+        if args:
+            lines.append(f"(positional args: {args!r})")
+
+        if kwargs.get("instructions") is not None:
+            lines.append("\n--- instructions ---")
+            lines.append(str(kwargs["instructions"]))
+
+        inp = kwargs.get("input")
+        if isinstance(inp, list):
+            for m in inp:
+                if isinstance(m, dict):
+                    lines.append(f"\n--- input[{m.get('role', '?')}] ---")
+                    lines.append(str(m.get("content", m)))
+                else:
+                    lines.append("\n--- input[?] ---")
+                    lines.append(str(m))
+        elif inp is not None:
+            lines.append("\n--- input ---")
+            lines.append(str(inp))
+        lines.append("=" * 72)
+        printer("\n".join(lines))
+
+        if not send:
+            raise RuntimeError("spy_openai(send=False): stopped before sending")
+
+        resp = orig(*args, **kwargs)
+        if show_response:
+            printer(f"--- response.output_parsed ---\n"
+                    f"{getattr(resp, 'output_parsed', resp)}\n{'=' * 72}")
+        return resp
+
+    CLIENT.responses.parse = _wrapper
+    try:
+        yield
+    finally:
+        if had_own_parse:
+            CLIENT.responses.parse = orig
+        else:
+            # Remove our instance-attribute shadow so the original class
+            # method shows through again.
+            CLIENT.responses.__dict__.pop("parse", None)
 
 
 def contains_i_am(x):
--- a/vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
+++ b/vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
@@ -77,6 +77,7 @@ the same data as a human-readable Markdown report.
 """
 from __future__ import annotations
 
+import hashlib
 from typing import Any, Iterable
 
 import pandas as pd
@@ -369,13 +370,20 @@ def check_taxonomy_coherence(
 
     cache = CoherenceAuditCache(session=session, model=model)
 
-    # Build (given_id, user_text) pairs. given_id is taxonomy-agnostic —
-    # the system prompt + AuditResponse schema (captured in prompt_id)
-    # already distinguish this audit from other tools; level + name
-    # uniquely identify a parent within one taxonomy.
+    # Build (given_id, user_text) pairs. The system prompt + AuditResponse
+    # schema (captured in prompt_id) distinguish this audit from other
+    # tools. Within one taxonomy, level + name is NOT unique — the same
+    # node name can recur under different parents (e.g. "Buildings" under
+    # two pillars), and those parents have different children. Since
+    # ``bulk_get_cache_or_run`` returns a dict keyed by given_id alone,
+    # a bare "level|name" key would let two such parents collide and share
+    # one response. Append a short hash of the user prompt (parent
+    # definition + children) so distinct parents get distinct keys; two
+    # parents with genuinely identical audits still collapse (correct),
+    # and an unchanged taxonomy re-runs to a full cache hit because the
+    # hash is content-derived.
     requests: list[tuple[str, str]] = []
     for t in tasks:
-        given_id = f"{t['parent_level']}|{t['parent_name']}"
         user_text = _build_user_prompt(
             parent_level=t["parent_level"],
             child_level=t["child_level"],
@@ -383,6 +391,8 @@ def check_taxonomy_coherence(
             parent_definition=t["parent_definition"],
             children=t["children"],
         )
+        content_hash = hashlib.sha1(user_text.encode("utf-8")).hexdigest()[:8]
+        given_id = f"{t['parent_level']}|{t['parent_name']}|{content_hash}"
         requests.append((given_id, user_text))
 
     # Restore deterministic audits (temperature=0); reasoning models
--- a/vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
+++ b/vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
@@ -66,9 +66,11 @@ Library entry point::
 All OpenAI calls flow through the SQL prompt/response cache
 (``CoherenceAuditCache`` extends ``InstructorPRC``). Cache keys: the
 fixed system prompt + the ``AuditResponse`` schema -> ``prompt_id``;
-``f"{parent_level}|{parent_name}"`` -> ``given_id``; hash of the user
-prompt body (parent definition + children) -> ``text_id``. Re-running
-the audit against an unchanged taxonomy is a no-API run.
+``f"{parent_level}|{parent_name}|{hash(user prompt body)}"`` ->
+``given_id`` (the content hash disambiguates same-named parents that
+recur at one level under different parents); hash of the user prompt
+body (parent definition + children) -> ``text_id``. Re-running the
+audit against an unchanged taxonomy is a no-API run.
 
 The output frame has one row per parent node with: parent level/name,
 n_children, verdict (ok/minor_gaps/major_gaps), summary, coverage_gaps,
--- a/vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
+++ b/vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
@@ -84,6 +84,7 @@ from pydantic import BaseModel
 
 from vdl_tools.shared_tools.openai.prompt_response_cache_instructor import InstructorPRC
 from vdl_tools.shared_tools.openai.prompt_response_cache_sql import DEFAULT_MODEL
+from vdl_tools.shared_tools.openai.openai_api_utils import is_reasoning_model
 
 
 # ---------------------------------------------------------------------------
@@ -280,6 +281,7 @@ def check_taxonomy_coherence(
     definition_col: str = "Definition",
     read_from_cache: bool = True,
     write_to_cache: bool = True,
+    temperature: float | None = 0,
 ) -> pd.DataFrame:
     """Audit parent-child coherence at every non-leaf level transition.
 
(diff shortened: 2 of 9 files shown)


vdl_tools/shared_tools/taxonomy_mapping/oe_hierarchical_taxonomy_mapping.py
```diff
@@ -915,54 +1089,63 @@ def map_to_oneearth(
...
-    )
-    if recover_unmatched:
-        # Default top-level column + category choices from the level spec +
-        # tables; override only the scope prompt with the OE-specific rich one.
-        per_row_df = _htm.recover_unmatched(
-            per_row_df,
-            client=client,
-            model=recovery_model,
+    with get_session(session=session) as session:
+        per_row_df = classify_entities(
```
ztawilvdl, line 1119: **[correctness] Model switch silently reuses the prior model's cached classifications.** `TaxonomyMatchCache` defaults `filter_by_model=False`, so cache identity is `(prompt_id, given_id, text_id)` with no model name. `classify_entities` exposes the flag (and its docstring warns to set it True across models), but `map_to_oneearth` neither accepts nor forwards it. Since this PR exists to A/B-compare `gpt-5.4-nano` vs `gpt-4.1`, the intended workflow is the one that corrupts: run model A, re-run model B on the same taxonomy+prompt in the same DB, and B serves A's rows — an unauditable comparison. Add a `filter_by_model` param to `map_to_oneearth` and forward it (consider defaulting it True).

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -679,267 +788,171 @@ def classify_entity(
...
-                    leaves.append(new_branch)
-                else:
-                    next_active.append(new_branch)
+                if not matches:
+                    if branch["leaf"] is not empty_leaf:
+                        leaves.append(branch)
+                    continue
+
+                for i, m in enumerate(matches):
+                    cand_row = candidates[candidates[key_col] == m["name"]].iloc[0]
```
ztawilvdl: **[correctness] One unhandled exception now aborts the entire batch.** The old `_classify_one` wrapped each entity's walk in try/except, returning `[]` so a bad entity got a null row and the run continued. Phase C only guards `model_validate_json`; this descent block (`cand_row = candidates[...].iloc[0]`, `_clean_matches`, branch spawning) has no per-entity isolation. Any unexpected exception propagates out of `classify_entities`, killing every remaining entity *after* the paid bulk API call already ran and discarding uncommitted in-memory results. Wrap per-branch processing so a single failure degrades to a null row.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -572,105 +538,248 @@ def call_openai_match(
...
+    initial_branches: list[dict[str, Any]] = []
+    base_by_entity_id: dict[Any, dict[str, Any]] = {}
+    for _, row in entities.iterrows():
+        entity_id = row[id_col]
+        name = str(row[name_col])
+        desc = str(row[text_col])
+        base = {col: row[col] for col in row.index}
+        base[name_col] = name
+        base[text_col] = desc
+        base_by_entity_id[entity_id] = base
```
ztawilvdl, line 763: **[correctness] Duplicate `id_col` rows are silently collapsed.** `base_by_entity_id[entity_id] = base` overwrites on duplicate ids, and two rows sharing an id build the same level-0 `given_id` (`f"{id}|{level}|"`) with different text; `bulk_get_cache_or_run` returns a dict keyed by `given_id` only, so both branches read one response. The old per-row iteration made no uniqueness assumption. Either de-dup/validate `id_col` up front or key by row index.

vdl_tools/shared_tools/taxonomy_mapping/oe_hierarchical_taxonomy_mapping.py
```diff
@@ -915,54 +1089,63 @@ def map_to_oneearth(
...
-        emit_per_level=emit_per_level,
-    )
-    if recover_unmatched:
-        # Default top-level column + category choices from the level spec +
-        # tables; override only the scope prompt with the OE-specific rich one.
-        per_row_df = _htm.recover_unmatched(
-            per_row_df,
-            client=client,
-            model=recovery_model,
+    with get_session(session=session) as session:
```
ztawilvdl: **[correctness/altitude] Re-wrapping a caller-passed session in `get_session` commits+closes a session you don't own.** The documented usage is `with get_session() as session: map_to_oneearth(session=session)`. Internally `with get_session(session=session)` re-enters a context manager whose `finally` unconditionally `session.close()`s (and force-commits) — even a session it didn't create. It doesn't hard-crash today only because SQLAlchemy tolerates commit/reuse after close, but it closes a session the caller still owns and breaks single-transaction scope. Use a passed-in session directly; wrap in `get_session` only when `session is None`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -679,267 +788,171 @@ def classify_entity(
...
-                entity_description=entity_description,
-                level_name=lvl["name"],
-                candidates=candidates,
-                key_col=key_col,
-                confidence_threshold=confidence_threshold,
+            user_text = _build_user_text(
+                branch["entity_name"], branch["entity_description"],
+                lvl["name"], candidates, key_col,
+            )
+            given_id = _build_given_id(
```
ztawilvdl, line 808: **[robustness/altitude — correction] `given_id` cache-key encoding is safe only by convention.** *(Revising my original framing, which was wrong.)* `candidates_for_level` is a pure function of `(level, parent_path)`, and so is `text` — same inputs as `given_id`. So two branches that reach a level with the same `parent_path` get identical candidates and identical text: a true duplicate that `bulk_get_cache_or_run` collapses correctly. **There is no bug from two parents sharing a filter value** — I was wrong about that. The only residual risk is separator injection in `path = ">".join(f"{k}={v}" ...)`: two structurally different `parent_path` dicts could serialize to the same string if a taxonomy value contains `>` or `=`. The docstring leans on an unenforced assumption that names never contain those characters. Low likelihood; cheap defensive fix (hash the sorted items, or assert names are separator-free). Downgrading this from correctness to a robustness nit.

vdl_tools/shared_tools/taxonomy_mapping/check_taxonomy_coherence.py
```diff
@@ -325,38 +364,81 @@ def check_taxonomy_coherence(
...
+
+    cache = CoherenceAuditCache(session=session, model=model)
+
+    # Build (given_id, user_text) pairs. given_id is taxonomy-agnostic —
+    # the system prompt + AuditResponse schema (captured in prompt_id)
+    # already distinguish this audit from other tools; level + name
+    # uniquely identify a parent within one taxonomy.
+    requests: list[tuple[str, str]] = []
+    for t in tasks:
+        given_id = f"{t['parent_level']}|{t['parent_name']}"
```
ztawilvdl: **[correctness] Coherence `given_id` collision mislabels audit rows — more likely than a rare edge case.** `given_id = f"{parent_level}|{parent_name}"` encodes level + name only, not where the node sits in the tree, but `text` is built from the parent’s definition **and its children**. If a node name recurs at one level under different parents (common: `"Buildings"`, `"Other"`, `"Policy"`, `"Cross-Cutting"` appearing under multiple pillars), both audit tasks share one `given_id` with different `text`. `bulk_get_cache_or_run` dedups on `(given_id, text)` so it makes both API calls (you pay twice), but the returned dict is keyed by `given_id` alone, so the second answer overwrites the first. The `zip(tasks, requests)` reconstruction then hands **both** parents the same verdict — one is audited correctly, the other reported with the wrong node’s coverage/completeness gaps, silently. Fix: make `given_id` encode the full ancestor path (or a hash of the child set), not just `level|name`.

vdl_tools/shared_tools/taxonomy_mapping/oe_hierarchical_taxonomy_mapping.py
```diff
@@ -900,11 +1056,29 @@ def map_to_oneearth(
...
+        confidence_threshold is not None and prompt_mode == "organization"
+    )
+    match_schema = oneearth_match_schema(
+        include_confidence=confidence_on,
+        research=(prompt_mode == "research"),
+    )
+
     if system_prompt is None:
-        if confidence_threshold is not None and prompt_mode == "organization":
+        if confidence_on:
```
ztawilvdl, line 1099: **[simplify/footgun] Dead `include_confidence` branch produces an identical prompt.** `build_oneearth_system_prompt(include_confidence=True)` no longer affects the prose (per its own docstring), so this branch yields a string identical to `_PROMPT_BY_MODE["organization"]`. The block collapses to `_PROMPT_BY_MODE[prompt_mode]`. Worse than dead: a maintainer who later re-diverges the prompt on confidence silently changes `prompt_id` and invalidates every cached row.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -572,105 +538,248 @@ def call_openai_match(
...
+    out_cols = [lvl["output_col"] for lvl in levels]
+    leaf_keys = ["deepest_match", "leaf_definition", "mode_of_operation",
+                 "evidence", "reason", "confidence"]
+    empty_leaf = {k: None for k in leaf_keys}
+    response_model: type[BaseModel] = match_schema or MatchesResponse
+
+    # Reasoning models reject `temperature`; suppress it for them.
+    # Otherwise default to 0 (legacy behavior, deterministic outputs).
+    api_kwargs: dict[str, Any] = {}
+    if temperature is not None and not is_reasoning_model(model):
```
ztawilvdl, line 737: **[reuse/altitude] Reasoning-model temperature suppression copy-pasted into 4 sites** (`classify_entities`, `recover_unmatched`, `check_taxonomy_coherence`, `run_judge_evaluation`). `InstructorPRC.get_completion` already special-cases reasoning models in one place — that's where "drop temperature for gpt-5*" belongs, so callers can pass `temperature=0` unconditionally. A new reasoning-model family otherwise means editing four copies in lockstep.

vdl_tools/shared_tools/openai/openai_api_utils.py
```diff
@@ -179,6 +180,107 @@ def get_completion(
     return response.output_parsed
 
 
+@contextlib.contextmanager
+def spy_openai(send: bool = True, printer=print, show_response: bool = True):
```
ztawilvdl, line 184: **[conventions/unused] Debug helper with zero callers added to a shared production module.** `spy_openai` (~100 lines) is never called repo-wide and monkey-patches the module-global `CLIENT.responses.parse` (not thread-safe if ever entered during concurrent cache workers). Relatedly, the new `__main__` block in `oe_hierarchical_taxonomy_mapping.py` hardcodes `taxonomy_dir=Path("../shared-data-clean/...")` (cwd-relative, reintroducing the machine-specific-path smell this PR removed) and defaults `write_to_cache=True`, so running the file spends live API budget. Both trip the "flag unused code" rule — consider dropping them or moving to a test/scratch module.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -679,267 +788,171 @@ def classify_entity(
...
+        # Phase A — assemble requests; carry seeded branches forward
+        # until they reach their start_idx.
+        requests: list[tuple[str, str, dict[str, Any], pd.DataFrame]] = []
         next_active: list[dict[str, Any]] = []
-
         for branch in active:
+            if level_idx < branch["start_idx"]:
+                next_active.append(branch)
+                continue
             candidates = candidates_for_level(tables, level_idx, branch["parent_path"])
```
ztawilvdl, line 799: **[efficiency] Candidate frame + matched row re-derived per branch/match.** `candidates_for_level(...)` re-filters and `reset_index` once per branch (N identical empty-filter copies at level 0), and at line 854 `cand_row = candidates[candidates[key_col] == m["name"]].iloc[0]` is an O(n) scan even though `_clean_matches` already resolved the row via `iloc[idx-1]` and discarded the index. Memoize the frame per `(level_idx, parent_path)` and carry the 0-based index through `_clean_matches`. Main-thread overhead only, but real for large hot runs.

ztawilvdl (commented):
## Automated code review — SQL-cache migration

High-effort review (8 finder angles, recall-biased verify). The refactor is well-structured and the schema-on-Pydantic design is sound; below are 10 findings, correctness first. Nothing here is a merge blocker on its own, but #1 (model-scoped cache), #2 (batch-abort isolation), and #4 (session ownership) are worth addressing before this drives a real A/B model comparison.

_Checked-and-dropped (not bugs): the `self.session = session` fallback removal is a net fix (the old `or get_session()` assigned a context-manager object, never a session); the `recover_unmatched` "dropped pillar/category key" concern is moot under structured output (the model must emit `ScopeDecision.category`); the `_clean_matches` non-int-index "crash" can't happen (Pydantic validates `index: int` first)._

PR #148, opened 2026-07-03
Add generic post-classification layer to hierarchical taxonomy engine

ztawilvdl:
## What

Adds a generic post-processing layer on top of `classify_entities` in `hierarchical_taxonomy_mapping.py`, so project drivers can turn per-row classifier output into the legacy mapping-column schema + a distributed-funding frame without reimplementing any of it. All new functions are expressible purely in terms of `(per_row, levels, id_col, name_col, mapping_name)` — no domain knowledge.

New functions:
- **`select_primary_paths`** — one "primary" match row per id. `strategy="deepest"` (deepest non-empty level, first on ties) now; accepts a callable so an LLM primary-picker (the hierarchical successor of `run_primary_category_selection`) can slot in later without signature churn.
- **`attach_mapping_columns`** — hierarchical successor of `taxonomy_mapping.add_mapping_to_orgs`. Same suffix convention (`mapping_name=None` → `level0`/`all_level0`/`mapped_category`; `"ed_category"` → `level0_ed_category`/…). Aggregate `all_levelN_*` columns are **native Python lists** (downstream commonly checks `isinstance(x, list)`); `pad_to_n_levels` emits blank stubs when a consumer expects a deeper legacy schema than the taxonomy has.
- **`distribute_funding_from_matches`** — exact-contract adapter onto the **unchanged** `redistribute_funding_fracs`, i.e. the same single call that produced `distr_df` inside the embedding pipeline's `add_taxonomy_mapping`. `FundingFrac` never depended on embedding scores — only on the set of matched paths — so equal-split / `No_Level_n` backfill / per-org normalization semantics are identical.
- **`add_hierarchical_taxonomy_mapping`** — orchestrator mirroring `add_taxonomy_mapping`'s return contract: `(df_with_columns, distr_df | None, per_row)`. Performs no file I/O — callers decide where frames land, so lineage stays with the project.

## Why

`ed_tracker`'s `process_enrich` is migrating off the embedding-based `add_lstudio_taxonomy` onto the hierarchical LLM classifier (see the paired ed_tracker PR). The classify → attach columns → select primary → distribute funding composition is a **cross-project need** (US Climate did it via `add_taxonomy_mapping`; One Earth will want it too), so it belongs in vdl-tools rather than any single driver.

## Notes for reviewers

- `classify_entities` / `collapse_to_one_row_per_uid` / `redistribute_funding_fracs` are **unchanged**.
- `distribute_funding_from_matches` imports `taxonomy_mapping` lazily (it pulls in the embedding stack this module otherwise doesn't need).
- Verified with synthetic `per_row` fixtures (multi-match, shallow-match, no-match orgs): native-list aggregates, deepest-match primary, `FundingFrac` sums to 1.0 per id, `No_Level_n` backfill, `pad_to_n_levels` stubs, and no list-aliasing across rows.
- First consumer is ed_tracker; the ed_tracker PR branch imports from this branch, so this should merge first.

### DIFF
--- a/vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
+++ b/vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
@@ -1067,8 +1067,12 @@ def collapse_to_one_row_per_uid(
 # (``taxonomy_mapping.add_mapping_to_orgs`` + the ``distr_df`` half of
 # ``add_taxonomy_mapping``), operating on ``classify_entities``' per-row
 # output instead of embedding matches. Everything here is expressible in
-# terms of (per_row, levels, id_col, name_col, mapping_name) — no domain
-# knowledge — so project drivers stay thin.
+# terms of (per_row_mapping_df, levels, id_col, name_col, mapping_name) —
+# no domain knowledge — so project drivers stay thin.
+#
+# ``per_row_mapping_df`` throughout is the DataFrame returned by
+# ``classify_entities``: one row per (entity, matched leaf path), plus one
+# all-null row per entity the walk left unmatched.
 
 def _path_depth(row, output_cols: list[str]) -> int:
     """0-based index of the deepest non-empty level value in a per-row record.
@@ -1085,33 +1089,34 @@ def _path_depth(row, output_cols: list[str]) -> int:
 
 
 def select_primary_paths(
-    per_row: pd.DataFrame,
+    per_row_mapping_df: pd.DataFrame,
     levels: list[dict],
     id_col: str,
     strategy="deepest",
 ) -> pd.DataFrame:
-    """Pick one "primary" match row per id from the per-row output.
+    """Pick one "primary" match row per id from the per-row mapping frame.
 
     ``strategy="deepest"`` keeps the row whose deepest non-empty level is
     deepest (first row on ties — per-row order is walk order, so this is
     deterministic). A callable can be passed instead — it receives
-    ``(per_row, levels, id_col)`` and must return the same shape — which is
-    the hook where an LLM primary-picker (the hierarchical successor of
-    ``run_primary_category_selection``) can slot in later.
+    ``(per_row_mapping_df, levels, id_col)`` and must return the same shape
+    — which is the hook where an LLM primary-picker (the hierarchical
+    successor of ``run_primary_category_selection``) can slot in later.
 
     Returns one row per id with columns ``[id_col, *output_cols,
     "cat_level"]`` where ``cat_level`` is the primary path's 0-based depth
     (pd.NA for ids with no match; their level values are all null).
     """
     if callable(strategy):
-        return strategy(per_row, levels, id_col)
+        return strategy(per_row_mapping_df, levels, id_col)
     if strategy != "deepest":
         raise ValueError(f"Unknown primary strategy {strategy!r}; expected 'deepest' or a callable")
 
     levels = normalize_levels(levels)
     output_cols = [lvl["output_col"] for lvl in levels]
 
-    work = per_row[[id_col] + [c for c in output_cols if c in per_row.columns]].copy()
+    present = [c for c in output_cols if c in per_row_mapping_df.columns]
+    work = per_row_mapping_df[[id_col] + present].copy()
     for c in output_cols:
         if c not in work.columns:
             work[c] = None
@@ -1126,7 +1131,7 @@ def select_primary_paths(
 
 def attach_mapping_columns(
     df: pd.DataFrame,
-    per_row: pd.DataFrame,
+    per_row_mapping_df: pd.DataFrame,
     levels: list[dict],
     id_col: str,
     mapping_name: str | None = None,
@@ -1158,15 +1163,15 @@ def attach_mapping_columns(
     suffix = f"_{mapping_name}" if mapping_name else ""
     category_col = mapping_name if mapping_name else "mapped_category"
 
-    grouped = per_row.groupby(id_col, sort=False)
+    grouped = per_row_mapping_df.groupby(id_col, sort=False)
     all_maps: dict[int, dict] = {}
     for i, col in enumerate(output_cols):
-        if col in per_row.columns:
+        if col in per_row_mapping_df.columns:
             all_maps[i] = grouped[col].apply(lambda s: _dedup_preserve(s.tolist())).to_dict()
         else:
             all_maps[i] = {}
 
-    primary_df = select_primary_paths(per_row, levels, id_col, strategy=primary)
+    primary_df = select_primary_paths(per_row_mapping_df, levels, id_col, strategy=primary)
     primary_df = primary_df.set_index(id_col)
 
     def _value_at_depth(r):
@@ -1193,13 +1198,13 @@ def attach_mapping_columns(
 
 
 def distribute_funding_from_matches(
-    per_row: pd.DataFrame,
+    per_row_mapping_df: pd.DataFrame,
     levels: list[dict],
     id_col: str,
     name_col: str,
     max_level: int = 2,
 ) -> pd.DataFrame:
-    """Build the per-(org, path) ``FundingFrac`` frame from per-row output.
+    """Build the per-(org, path) ``FundingFrac`` frame from the mapping frame.
 
     Exact-contract adapter onto ``redistribute_funding_fracs`` — the same
     single call that produced ``distr_df`` inside the embedding pipeline's
@@ -1225,8 +1230,8 @@ def distribute_funding_from_matches(
     levels = normalize_levels(levels)
     output_cols = [lvl["output_col"] for lvl in levels]
 
-    depths = per_row.apply(lambda r: _path_depth(r, output_cols), axis=1)
-    matched = per_row[depths >= 0]
+    depths = per_row_mapping_df.apply(lambda r: _path_depth(r, output_cols), axis=1)
+    matched = per_row_mapping_df[depths >= 0]
     paths_df = pd.DataFrame({
         id_col: matched[id_col],
         name_col: matched[name_col],
@@ -1252,24 +1257,34 @@ def distribute_funding_from_matches(
 
 def add_hierarchical_taxonomy_mapping(
     *,
-    session,
     df: pd.DataFrame,
-    tables: dict[int, pd.DataFrame],
     levels: list[dict],
-    system_prompt: str,
     id_col: str,
     name_col: str,
-    text_col: str,
-    match_schema: type[BaseModel] | None = None,
-    model: str = "gpt-4.1",
-    descent_fanout_cap: int = 3,
-    max_workers: int = 8,
     mapping_name: str | None = None,
     primary="deepest",
     distribute_funding: bool = True,
     max_distr_funding_level: int = 2,
     funding_name_col: str | None = None,
     pad_to_n_levels: int | None = None,
+    # --- classification inputs -------------------------------------------
+    # Either pass a precomputed ``per_row_mapping_df`` (skip classification
+    # entirely — the escape hatch for drivers with a custom classify wrapper,
+    # e.g. drawdown's scope-recovery ``classify_entities_extended``), or
+    # supply the inputs below and this orchestrates the classify call.
+    # ``classify_fn`` is called with the standard ``classify_entities``
+    # kwargs, so an injected wrapper must accept that signature (or bake its
+    # extras in via ``functools.partial``).
+    per_row_mapping_df: pd.DataFrame | None = None,
+    classify_fn=classify_entities,
+    session=None,
+    tables: dict[int, pd.DataFrame] | None = None,
+    system_prompt: str | None = None,
+    text_col: str | None = None,
+    match_schema: type[BaseModel] | None = None,
+    model: str = "gpt-4.1",
+    descent_fanout_cap: int = 3,
+    max_workers: int = 8,
     read_from_cache: bool = True,
     write_to_cache: bool = True,
     temperature: float | None = 0,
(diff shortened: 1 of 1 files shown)


vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+    if paths_df.empty:
+        logger.warning("distribute_funding_from_matches: no matched paths")
+        paths_df["FundingFrac"] = pd.Series(dtype=float)
+        return paths_df
+
+    return redistribute_funding_fracs(
+        paths_df,
+        id_attr=id_col,
+        keepcols=[name_col],
+        max_level=max_level,
```
ztawilvdl, line 1296: **Default `max_distr_funding_level=2` silently drops the funding frame for any taxonomy with fewer than 3 levels.**

`redistribute_funding_fracs` returns `None` (not a DataFrame) when the taxonomy is shallower than `max_level` — `if top_lev < max_level: logger.error(...); return None` (taxonomy_mapping.py:345-347). A 2-level taxonomy yields `output_cols = [level0, level1]`, so `top_lev = 1 < 2`.

Result: `distribute_funding_from_matches` returns `None`, and `add_hierarchical_taxonomy_mapping` returns `(df, None, per_row_df)` even though `distribute_funding=True` and matches exist. `None` is also the `distribute_funding=False` sentinel, so callers cannot distinguish "disabled" from "silently failed" — a downstream `distr.to_parquet(...)` / `len(distr)` crashes far from the cause. Derive the level from the taxonomy, e.g. `max_level=min(max_distr_funding_level, len(output_cols) - 1)`, or raise when the requested depth exceeds the taxonomy.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+    present = [c for c in output_cols if c in per_row_mapping_df.columns]
+    work = per_row_mapping_df[[id_col] + present].copy()
+    for c in output_cols:
+        if c not in work.columns:
+            work[c] = None
+    work["_depth"] = work.apply(lambda r: _path_depth(r, output_cols), axis=1)
+
+    # idxmax keeps the first occurrence of the max depth per id.
+    idx = work.groupby(id_col, sort=False)["_depth"].idxmax()
+    primary = work.loc[idx, [id_col] + output_cols + ["_depth"]].reset_index(drop=True)
```
ztawilvdl, line 1139: **`select_primary_paths` assumes a unique index on `per_row_mapping_df`; the documented escape hatch can violate it.**

`groupby(id_col)["_depth"].idxmax()` returns index *labels*, and `work.loc[idx, ...]` then selects **all** rows sharing each label. `classify_entities` builds a fresh `RangeIndex` (unique), but the documented escape hatch — passing a precomputed frame, e.g. drawdown's `classify_entities_extended` recovery frame built via `pd.concat` without `reset_index` — can carry a non-unique index.

Verified locally: with a duplicated index, `idxmax` + `.loc[idx]` returns 3 rows where 2 are expected. The duplicated "primary" rows then flow into `attach_mapping_columns`, whose `primary_df.set_index(id_col)` becomes non-unique, and `ids.map(primary_df[col])` raises `InvalidIndexError: Reindexing only valid with uniquely valued Index objects`. Defensively `reset_index(drop=True)` the input (or the `work` frame) before `idxmax`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+    # this module otherwise doesn't need.
+    from vdl_tools.shared_tools.taxonomy_mapping.taxonomy_mapping import (
+        redistribute_funding_fracs,
+    )
+
+    levels = normalize_levels(levels)
+    output_cols = [lvl["output_col"] for lvl in levels]
+
+    depths = per_row_mapping_df.apply(lambda r: _path_depth(r, output_cols), axis=1)
+    matched = per_row_mapping_df[depths >= 0]
```
ztawilvdl, line 1274: **Non-unique-index input also corrupts the funding frame here (silently, not a crash).**

`depths = per_row_mapping_df.apply(..., axis=1)` shares the input's index; `per_row_mapping_df[depths >= 0]`, `matched[name_col]`, and `depths[depths >= 0]` are then aligned by index inside the `pd.DataFrame({...})` construction. If the precomputed `per_row_mapping_df` (escape hatch) has a non-unique index, the boolean-mask selections and the positional column assembly misalign, so `id_col` / `name_col` / `cat_level` in a `paths_df` row can come from different source rows → corrupted per-org funding distribution with no error. Same root cause as the `select_primary_paths` finding — a single `reset_index(drop=True)` at the top of both functions fixes both.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+                ("system_prompt", system_prompt), ("text_col", text_col),
+            ) if v is None
+        ]
+        if missing:
+            raise ValueError(
+                "add_hierarchical_taxonomy_mapping needs "
+                f"{missing} to classify — supply them, or pass a precomputed "
+                "per_row_mapping_df."
+            )
+        per_row_mapping_df = classify_fn(
```
ztawilvdl, line 1381: **`filter_by_model` (and `confidence_threshold` / `emit_per_level` / `seed_col`) can't be threaded through the orchestrator, so multi-model runs silently reuse another model's cache.**

`classify_entities` warns in its own docstring: *"Set `filter_by_model` True whenever you run more than one model against the same taxonomy, otherwise a model switch silently reuses the prior model's cached results."* But `add_hierarchical_taxonomy_mapping` exposes `model=...` while never forwarding `filter_by_model`, so it's stuck at the unsafe default `False`. A caller that runs `gpt-5-nano` then `gpt-4.1` through this orchestrator gets the first model's cached matches back for the second run — a lineage/correctness violation they can't fix without bypassing the function. Forward `filter_by_model` (at minimum) through to `classify_fn`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+        if pd.isna(d):
+            return None
+        return r[output_cols[int(d)]]
+
+    primary_category = primary_df.apply(_value_at_depth, axis=1)
+
+    ids = df[id_col]
+    for i, col in enumerate(output_cols):
+        # `.get(x, ())` + list() yields a fresh list per row — no aliasing.
+        df[f"all_level{i}{suffix}"] = ids.map(lambda x, i=i: list(all_maps[i].get(x, ())))
```
ztawilvdl, line 1216: **`id_col` dtype skew between `df` and `per_row_mapping_df` silently blanks every mapping column.**

`all_maps[i]` is keyed by `per_row_mapping_df[id_col]` values and `primary_df` is indexed by them, but the columns are attached via `ids.map(...)` where `ids = df[id_col]`. If the two frames carry the same logical ids in different dtypes (e.g. `df` has Python `str "123"` while `classify_entities` coerced ids to `int64`, or one side is object-wrapped), every `.get(x, ())` / index lookup misses. Verified locally: `pd.Series(["1"]).map(lambda x: {1:...}.get(x,"MISS"))` → `"MISS"`. Result: `all_level{i}` uniformly `[]` and `level{i}`/`mapped_category` uniformly null for **all** rows, with no error. Consider asserting matching dtypes, or normalizing `id_col` on both frames before the map.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+        df[f"level{i}{suffix}"] = ids.map(primary_df[col])
```
ztawilvdl: **Whitespace normalization is inconsistent across the emitted columns.**

`level{i}` here (and `mapped_category` via `_value_at_depth`, line 1181) copy `primary_df[col]` / the raw `output_col` value **unstripped**, while `all_level{i}` uses `_dedup_preserve` (which does `str(v).strip()`) and `distribute_funding_from_matches` strips before building the funding frame (line 1242). So a taxonomy/leaf value with trailing whitespace becomes `"Solar "` in `level2`/`mapped_category` but `"Solar"` in `all_level2` and in `distr_df.level2`. A downstream join of `df.mapped_category` against `distr_df.level2`, or a `groupby(mapped_category)`, then misses or double-counts `"Solar"` vs `"Solar "`. Apply the same strip when writing `level{i}` and `mapped_category`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+    work = per_row_mapping_df[[id_col] + present].copy()
+    for c in output_cols:
+        if c not in work.columns:
+            work[c] = None
+    work["_depth"] = work.apply(lambda r: _path_depth(r, output_cols), axis=1)
+
+    # idxmax keeps the first occurrence of the max depth per id.
+    idx = work.groupby(id_col, sort=False)["_depth"].idxmax()
+    primary = work.loc[idx, [id_col] + output_cols + ["_depth"]].reset_index(drop=True)
+    primary["cat_level"] = primary["_depth"].map(lambda d: pd.NA if d < 0 else int(d))
```
ztawilvdl: **`cat_level` mixes `pd.NA` and `int` → object/nullable dtype that breaks integer consumers.**

`primary["cat_level"] = primary["_depth"].map(lambda d: pd.NA if d < 0 else int(d))` produces a column mixing `pd.NA` (no-match ids) with Python ints, so the attached `cat_level{suffix}` column is object/nullable dtype. Legacy consumers treat `cat_level` as a plain int (`df.cat_level != 0`, `rdf.cat_level > max_level`, `.astype(int)`). With `pd.NA` present, `.astype(int)` raises and `== 0` yields `pd.NA` (not `False`), so boolean indexing behaves unpredictably. The legacy `add_mapping_to_orgs` never carried no-match ids into `cat_level` at all. Consider a sentinel int (e.g. `-1`) or a nullable `Int64` dtype documented for consumers.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+
+    if paths_df.empty:
+        logger.warning("distribute_funding_from_matches: no matched paths")
+        paths_df["FundingFrac"] = pd.Series(dtype=float)
+        return paths_df
+
+    return redistribute_funding_fracs(
+        paths_df,
+        id_attr=id_col,
+        keepcols=[name_col],
```
ztawilvdl: **`keepcols=[name_col]` crashes when `funding_name_col` (or `name_col`) equals `id_col`.**

`redistribute_funding_fracs` does `entity_df = df.set_index(id_attr)[entity_attrs]` (taxonomy_mapping.py:500) where `entity_attrs = funding_attrs + keepcols`. If a driver passes `funding_name_col=id_col` (reusing the uid as the display name — a reasonable choice), `keepcols=[id_col]` and the `set_index(id_col)[[id_col]]` selection raises `KeyError` because `id_col` was consumed by `set_index` — aborting the funding stage *after* classification already spent API budget. Guard against `name_col == id_col`, or select keepcols before `set_index`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+
+    levels = normalize_levels(levels)
+    output_cols = [lvl["output_col"] for lvl in levels]
+
+    present = [c for c in output_cols if c in per_row_mapping_df.columns]
+    work = per_row_mapping_df[[id_col] + present].copy()
+    for c in output_cols:
+        if c not in work.columns:
+            work[c] = None
+    work["_depth"] = work.apply(lambda r: _path_depth(r, output_cols), axis=1)
```
ztawilvdl, line 1135: **`_path_depth` is recomputed row-by-row via `.apply(axis=1)` three times per `add_hierarchical_taxonomy_mapping` call.**

The per-row depth scan runs here (`select_primary_paths`), again in `distribute_funding_from_matches` (line 1233), and a third time because `attach_mapping_columns` calls `select_primary_paths`. Each is an object-dtype Python loop over the full per-row frame — the dominant post-processing cost on large classification frames, and `_value_at_depth` (line 1183) adds a fourth `apply(axis=1)`. Compute a single vectorized `_depth` Series once over `per_row_mapping_df[output_cols]` (reverse column scan) and thread it into both consumers; also emit the primary category value directly from `select_primary_paths` (it already knows the row and its depth) instead of re-deriving it with `_value_at_depth`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1060,6 +1060,311 @@ def collapse_to_one_row_per_uid(
...
+      - ``pad_to_n_levels`` — when consumers expect a deeper legacy schema
+        than the taxonomy has, emit blank singular (null) and list (``[]``)
+        columns for the missing levels.
+    """
+    levels = normalize_levels(levels)
+    output_cols = [lvl["output_col"] for lvl in levels]
+    suffix = f"_{mapping_name}" if mapping_name else ""
+    category_col = mapping_name if mapping_name else "mapped_category"
+
+    grouped = per_row_mapping_df.groupby(id_col, sort=False)
```
ztawilvdl, line 1194: **Per-id list aggregation + deepest-level detection duplicate `collapse_to_one_row_per_uid`, and run a second `groupby` over the same frame.**

`attach_mapping_columns` runs `grouped[col].apply(_dedup_preserve)` per level here, then calls `select_primary_paths`, which runs its *own* `groupby(id_col)` over the identical frame — two hash-partition passes. The per-id ordered-unique lists and the deepest-non-empty-level logic are exactly what `collapse_to_one_row_per_uid` (line 1001, via `_dedup_preserve` + its `deepest_name` walk) already computes. Two independent notions of "deepest match" now exist: if the blank/emptiness rule ever changes, `cat_level`/`FundingFrac` and the collapsed `deepest_match` silently diverge. Reuse a single `grouped = per_row_mapping_df.groupby(id_col, sort=False)` for both the lists and the primary selection, and factor the depth rule into one shared helper.

PR #152, opened 2026-07-09
Add json_cache with S3 local-filecache; extract shared cache backend

ztawilvdl:
## What

Generalizes the transparent local-caching machinery in `parquet_cache.py` so **JSON** gets the same easy local / `s3://` I/O — with remote reads cached locally and ETag-validated on every open (no silent stale reads).

Three files, **no new dependencies** (stdlib `json` + `gzip`, existing `fsspec`):

| File | What |
|---|---|
| `_s3_cache_backend.py` (new) | Shared S3 / creds / ETag-filecache / bucket-create core, extracted from `parquet_cache` so credential and caching behavior can't drift between modules. |
| `json_cache.py` (new) | `write_json` / `read_json` / `get_lineage` / `write_jsons` / `read_jsons`, mirroring the `parquet_cache` API and cache semantics (`use_cache` / `check_remote` / `cache_dir`). |
| `parquet_cache.py` (changed) | Rewired onto the shared backend. Public API and behavior unchanged. |

## Why

The valuable part of `parquet_cache` was never Parquet-specific — it's the `s3://` + local-`filecache`-with-ETag-validation plumbing. Extracting it lets JSON reuse the same "speed optimization" and gives us one code path for creds/caching instead of two that can drift.

## Design decisions for a reviewer

1. **Data files stay pure JSON.** Parquet hides lineage in the footer; JSON has none. Since our public buckets often serve JSON straight to frontends, the bytes on disk / in S3 are exactly what `json.dumps` produced. Optional `lineage` goes to a small sidecar (`{uri}.vdl.json`) written only when passed, read back via `get_lineage`; `read_json` never touches it.
2. **Compression by extension.** `…​.json` → raw (frontend-servable); `…​.json.gz` → gzipped (JSON compresses ~10×, good for internal blobs). Note: keep frontend-fetched files as `.json` — fsspec writes no `Content-Encoding` header, so browsers won't auto-decompress `.gz`.
3. **Refactored rather than duplicated.** Cache root unified under `~/.cache/vdl-tools/{parquet,json}`. `VDL_PARQUET_CACHE_DIR` is still honored; `VDL_JSON_CACHE_DIR` added.

## Testing

Local round-trip smoke test passed: raw JSON (pure bytes verified), `.json.gz` (gzip magic + lineage sidecar), multi-file `write_jsons`/`read_jsons`, and `parquet_cache` round-trip (incl. JSON columns + lineage) confirming the refactor is behavior-preserving. Verified no other modules import the private helpers that moved.

## Not included

No `prune_cache` for the JSON cache dir — `parquet_cache.prune_cache(cache_dir=...)` already works on any directory. Open to hoisting it into the shared backend if preferred.

### DIFF
--- a/vdl_tools/shared_tools/_s3_cache_backend.py
+++ b/vdl_tools/shared_tools/_s3_cache_backend.py
@@ -90,5 +90,10 @@ def write_target(uri: str) -> dict:
         if not bucket_exists(bucket):
             create_bucket(get_s3_client(), bucket)
         return {"s3": s3_creds()}
-    Path(uri).parent.mkdir(parents=True, exist_ok=True)
+    # Local target: ensure the parent dir exists. Strip a ``file://`` scheme
+    # first — ``Path("file:///a/b")`` is a *relative* path rooted at ``file:``,
+    # so ``.parent.mkdir`` would spray a junk ``file:`` tree into the cwd
+    # instead of creating the real parent.
+    local_path = uri[len("file://"):] if uri.startswith("file://") else uri
+    Path(local_path).parent.mkdir(parents=True, exist_ok=True)
     return {}
--- a/vdl_tools/shared_tools/_s3_cache_backend.py
+++ b/vdl_tools/shared_tools/_s3_cache_backend.py
@@ -1,0 +1,94 @@
+"""Shared S3 + local-filecache plumbing for the ``*_cache`` modules.
+
+This is the storage-agnostic core extracted from :mod:`parquet_cache`: URI
+detection, AWS credential resolution, the ETag-validated local filecache used
+for ``s3://`` reads, and bucket/parent-dir creation on write. Both
+:mod:`parquet_cache` and :mod:`json_cache` build on it so the credential and
+caching behaviour never drifts between them.
+
+Not a public API — import the ``parquet_cache`` / ``json_cache`` wrappers
+instead.
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+from vdl_tools.shared_tools.tools.config_utils import get_configuration
+from vdl_tools.shared_tools.s3_tools import get_s3_client, create_bucket, bucket_exists
+
+
+# Root for all vdl-tools caches; each module gets its own subdir underneath.
+DEFAULT_CACHE_ROOT = Path(
+    os.environ.get("VDL_CACHE_DIR", Path.home() / ".cache" / "vdl-tools")
+)
+
+
+def is_s3(uri: str) -> bool:
+    return uri.startswith("s3://")
+
+
+def bucket_of(uri: str) -> str:
+    """``s3://bucket/key`` -> ``bucket``."""
+    return uri.split("/")[2]
+
+
+def s3_creds() -> dict:
+    """Read AWS creds from config.ini ``[aws]``. Empty dict -> boto3 default chain."""
+    try:
+        aws = get_configuration()["aws"]
+    except Exception:
+        return {}
+    opts: dict = {}
+    if aws.get("access_key_id"):
+        opts["key"] = aws["access_key_id"]
+    if aws.get("secret_access_key"):
+        opts["secret"] = aws["secret_access_key"]
+    if aws.get("region"):
+        opts["client_kwargs"] = {"region_name": aws["region"]}
+    return opts
+
+
+def read_target(
+    uri: str,
+    use_cache: bool,
+    cache_dir: Path,
+    check_remote: bool,
+) -> tuple[str, dict]:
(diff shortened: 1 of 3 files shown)


vdl_tools/shared_tools/_s3_cache_backend.py
```diff
@@ -0,0 +1,94 @@
...
+
+    For ``s3://`` targets, create the bucket if it is missing. For local
+    targets, create the parent directory. Returns the fsspec ``open`` kwargs.
+    """
+    if is_s3(uri):
+        bucket = bucket_of(uri)
+        if not bucket_exists(bucket):
+            create_bucket(get_s3_client(), bucket)
+        return {"s3": s3_creds()}
+    Path(uri).parent.mkdir(parents=True, exist_ok=True)
```
ztawilvdl: **Bug: `file://` write targets create a junk `file:` directory tree in the cwd.**

`Path("file:///data/out/x.json").parent` is `file:/data/out` — a *relative* path rooted at `file:` (`is_absolute()` is `False`), so `.parent.mkdir(parents=True)` creates `./file:/data/out/` in the current working directory instead of the real parent. The actual write only succeeds because fsspec `LocalFileSystem` auto-creates parents; the mkdir here is redundant for plain local paths and actively wrong for the `file://` URIs this module documents as supported (it governs both data and lineage-sidecar writes). Reproduced: writing `file:///tmp/.../f.json` leaves a stray `file:` dir in cwd.

Suggested fix — strip the scheme before mkdir:

```suggestion
        return {"s3": s3_creds()}
    # Local target: ensure the parent dir exists. Strip a ``file://`` scheme
    # first — ``Path("file:///a/b")`` is a *relative* path rooted at ``file:``,
    # so ``.parent.mkdir`` would spray a junk ``file:`` tree into the cwd
    # instead of creating the real parent.
    local_path = uri[len("file://"):] if uri.startswith("file://") else uri
    Path(local_path).parent.mkdir(parents=True, exist_ok=True)
    return {}
```

vdl_tools/shared_tools/json_cache.py
```diff
@@ -0,0 +1,223 @@
...
+        f.write(data)
+
+    if lineage is not None:
+        sidecar = uri + _SIDECAR_SUFFIX
+        payload = json.dumps(
+            {"created_at": datetime.now(timezone.utc).isoformat(), **lineage},
+            default=str,
+        ).encode("utf-8")
+        # sidecar is always small + raw JSON (no compression)
+        with fsspec.open(sidecar, "wb", **_b.write_target(sidecar)) as f:
```
ztawilvdl: **Efficiency: the lineage sidecar re-runs `write_target()`, doubling the bucket check on every lineage-bearing S3 write.**

The data write at line 87 already called `write_target(uri)`, which for an `s3://` target issues a `bucket_exists()` HEAD (building a fresh boto3 client) and creates the bucket if missing. Calling `_b.write_target(sidecar)` here HEADs the *same* bucket a second time — the sidecar always lives in the bucket/dir the data write just ensured. Every `write_json(..., lineage=...)` to S3 pays two HEAD round-trips instead of one, and `write_jsons` fans this out across the batch.

Reuse the `opts` already computed above:

```suggestion
        # sidecar shares the data file's bucket / parent dir, so reuse ``opts``
        # rather than re-running write_target (which would re-issue a bucket_exists
        # HEAD, or re-create the parent, for a target already prepared above).
        # Always small + raw JSON (no compression).
        with fsspec.open(sidecar, "wb", **opts) as f:
            f.write(payload)
```

vdl_tools/shared_tools/json_cache.py
```diff
@@ -0,0 +1,223 @@
...
+def _compression(uri: str) -> str | None:
+    """gzip iff the path ends in ``.gz`` / ``.gzip`` — otherwise raw JSON."""
+    return "gzip" if uri.endswith((".gz", ".gzip")) else None
+
+
+# ---------------------------------------------------------------------------
+# Public API — single file
+# ---------------------------------------------------------------------------
+
+def write_json(
```
ztawilvdl, line 54: **Reuse note (low): overlaps existing `s3_tools.read_s3_json` / `write_json_to_s3`.**

`s3_tools.py` already exposes `read_s3_json()` (L304) and `write_json_to_s3()` (L330) for JSON-on-S3 I/O. This module is justified — it adds the ETag-validated local filecache and lineage sidecars those helpers lack — but the codebase now has two JSON-to-S3 paths with independent bucket/creds handling, which is the kind of drift this PR set out to eliminate for parquet/json. Per CLAUDE.md ("Use what we have first"), worth either a one-line docstring pointer distinguishing the two, or a follow-up to route the old helpers through this backend. Not blocking.

PR #153, opened 2026-07-10
Make NZI company search resumable, concurrent, and retry transient failures

ztawilvdl:
## What changed

`search_companies` fetched pages one at a time with no retries, holding all results in memory — a process dying at page 400 of 500 lost everything and restarted from page 1. This PR makes the search path resilient and concurrent, and speeds up the details phase.

### Resilience
- **Per-page search checkpoints** (`SearchCheckpoint` in `netzero_api.py`): each fetched page is persisted via `json_cache` (local path or `s3://`) under a directory keyed by a sha256 of the full query payload (endpoint + filter + sorting + page_size). Re-running the same query resumes from the completed pages; a changed query hashes to a different directory, so a checkpoint can never be resumed against the wrong query. The persisted payload doubles as a lineage record of exactly which query produced which results. Checkpoints older than 7 days are refetched (`checkpoint_max_age_days`).
- `search_companies` checkpoints **by default** to `~/.cache/vdl-tools/nzi_search` (override with `NZI_SEARCH_CHECKPOINT_DIR`; disable with `checkpoint_dir=None`).
- **Retries with backoff + jitter** for 429/5xx/connection errors on both the sync (`_request_with_retries`) and aiohttp (`_get_details_batch`) paths — previously only timeouts retried, and the search path had no retries at all.
- **Re-auth on 401/403**: the session cookie expires on multi-hour runs; both paths now re-login once and retry, with a lock + generation counter so N concurrent 401s cause exactly one re-login.

### Speed
- **Concurrent page fetches**: page 1 establishes `total_count`, remaining offsets fetch in parallel (default 8, `max_concurrent_pages`). Results are deduped by `clientID`/`investorID` in case rows shift between pages.
- **Details stages gathered**: `get_full_details_from_company_ids` runs startup details + funding rounds + commercial deals in one `asyncio.gather` instead of serial `asyncio.run` calls (investors still sequence after funding rounds, which they depend on).

### Bug fixes
- `max_pages = limit // page_size` truncated results (`limit=250` returned 200); now `ceil` + exact truncation.
- `UnboundLocalError` when `_search_entities` was called with no filter.
- Removed `_paginate` (superseded; only its own test used it).

## Tests

`test_netzero_api.py` rewritten — 6 of its 10 tests already failed on main against methods that no longer exist. Now 19 passing tests covering the checkpoint lifecycle (write, full resume with zero HTTP calls, partial resume, staleness, query isolation), 500-retry, 401 re-auth (sync + async), persistent-401 raising, dedupe, and the limit fix. A separate smoke test verified the gathered detail stages actually overlap and investors sequence after funding.

The 27 failures in `test_nzi_funding_round_processing.py` pre-date this branch (verified against the stashed tree) and are unrelated.

## Reviewer notes

- **No default sort order is applied.** I couldn't verify which sort field names the NZI API accepts without live calls, so a checkpointed search without explicit `sorting` logs a warning instead of guessing. Worth one live check (`Sorting(field="clientID")`) — if the API accepts it, we should make it the default for checkpointed searches.
- Unit tests mock all HTTP. Before a large production crawl, do one small live run to confirm the API tolerates 8 concurrent page fetches.
- Peak API concurrency for gathered detail stages is a small multiple of `max_concurrent_requests` (each stage gets its own semaphore); defaults kept conservative (10).

### DIFF
--- a/vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
+++ b/vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
@@ -1,13 +1,24 @@
 
-import os
-import requests
-import aiohttp
 import asyncio
-from typing import Dict, List, Optional, Union, Generator, Type
+import hashlib
+import json
+import math
+import os
+import random
+import threading
+import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from datetime import datetime, timedelta, timezone
+from typing import Dict, List, Optional, Type
+
+import aiohttp
+import requests
+
 from vdl_tools.shared_tools.tools.logger import logger
 from vdl_tools.scrape_enrich.netzero_insights.filters import (
     Sorting, MainFilter,
 )
+from vdl_tools.shared_tools.json_cache import read_json, write_json, target_exists
 
 from vdl_tools.shared_tools.database_cache.database_models import (
     CompanyCommercialDeal,
@@ -24,6 +35,78 @@ urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 PROD_BASE_URL = "https://api.netzeroinsights.com"
 SANDBOX_BASE_URL = "https://20.108.20.67"
 
+RETRYABLE_STATUS_CODES = frozenset({429, 500, 502, 503, 504})
+AUTH_STATUS_CODES = frozenset({401, 403})
+
+DEFAULT_SEARCH_CHECKPOINT_DIR = os.environ.get(
+    "NZI_SEARCH_CHECKPOINT_DIR",
+    os.path.expanduser("~/.cache/vdl-tools/nzi_search"),
+)
+
+
+class SearchCheckpoint:
+    """Per-page checkpoint store for a paginated search.
+
+    Pages are keyed by offset under a directory derived from a hash of the
+    full search payload, so a checkpoint can never be resumed against a
+    different query. Storage goes through ``json_cache`` (local path or
+    ``s3://``).
+    """
+
+    def __init__(self, base_dir: str, endpoint: str, payload: Dict, page_size: int):
+        self.payload = payload
+        self.page_size = page_size
+        self.endpoint = endpoint
+        key_material = json.dumps(
+            {"endpoint": endpoint, "payload": payload, "page_size": page_size},
+            sort_keys=True,
+            default=str,
+        )
+        self.search_key = hashlib.sha256(key_material.encode("utf-8")).hexdigest()[:16]
+        self.base_uri = f"{str(base_dir).rstrip('/')}/{self.search_key}"
+
+    def _page_uri(self, offset: int) -> str:
+        return f"{self.base_uri}/page_{offset:09d}.json.gz"
+
+    @property
+    def _meta_uri(self) -> str:
+        return f"{self.base_uri}/meta.json"
+
+    def load_meta(self, max_age: timedelta) -> Optional[Dict]:
+        """Return the checkpoint metadata, or None if absent or older than max_age."""
+        if not target_exists(self._meta_uri):
+            return None
+        meta = read_json(self._meta_uri)
+        created_at = datetime.fromisoformat(meta["created_at"])
+        if datetime.now(timezone.utc) - created_at > max_age:
+            logger.info(
+                "Search checkpoint %s is older than %s — refetching all pages",
+                self.base_uri, max_age,
+            )
+            return None
+        return meta
+
+    def write_meta(self, total_count: int) -> None:
+        write_json(
+            {
+                "created_at": datetime.now(timezone.utc).isoformat(),
+                "endpoint": self.endpoint,
+                "payload": self.payload,
+                "page_size": self.page_size,
+                "total_count": total_count,
+            },
+            self._meta_uri,
+        )
+
+    def has_page(self, offset: int) -> bool:
+        return target_exists(self._page_uri(offset))
+
+    def read_page(self, offset: int) -> Dict:
+        return read_json(self._page_uri(offset))
+
+    def write_page(self, offset: int, page: Dict) -> None:
+        write_json(page, self._page_uri(offset))
+
 
 class NetZeroAPI:
     """Client for interacting with the NetZero Insights API."""
@@ -34,7 +117,8 @@ class NetZeroAPI:
         password: str,
         use_sandbox: bool = False,
         read_from_cache: bool = True,
-        write_to_cache: bool = True
+        write_to_cache: bool = True,
+        max_concurrent_requests: int = 10,
     ):
         """Initialize the API client with credentials.
 
@@ -42,6 +126,10 @@ class NetZeroAPI:
             username: NetZero Insights API username
             password: NetZero Insights API password
             use_sandbox: Whether to use the sandbox environment
+            max_concurrent_requests: Cap on simultaneous requests per detail
+                batch. Note that stages fetched in parallel (e.g. startup
+                details + funding rounds) each get their own cap, so peak
+                concurrency against the API can be a small multiple of this.
         """
         logger.info(f"Initializing NetZero API client with {'sandbox' if use_sandbox else 'production'} environment")
         self.username = username
@@ -51,10 +139,13 @@ class NetZeroAPI:
         self.read_from_cache = read_from_cache
         self.write_to_cache = write_to_cache
         self.use_sandbox = use_sandbox
+        self.max_concurrent_requests = max_concurrent_requests
         # The production ssl certifcate seems to expired too
         self.verify_ssl = not use_sandbox
         # self.verify_ssl = False
 
+        self._auth_lock = threading.Lock()
+        self._auth_generation = 0
         self._authenticate()
 
     def _authenticate(self) -> None:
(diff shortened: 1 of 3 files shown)


vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -388,25 +565,47 @@ async def _get_details_batch(
...
                             headers={"Accept": "application/json, text/plain, */*"},
                             timeout=request_timeout,
                             ssl=self.verify_ssl,
                         ) as response:
+                            if response.status in AUTH_STATUS_CODES and not reauthed:
+                                # Session cookie likely expired mid-run; refresh
+                                # once (generation-guarded, so concurrent tasks
+                                # trigger a single re-login) and retry
+                                logger.warning(f"HTTP {response.status} for {endpoint} {id} — re-authenticating")
+                                auth_state["generation"] = await asyncio.to_thread(
```
ztawilvdl, line 613: **[correctness, high] Async re-auth generation dedup is defeated → re-login storm.** `fetch_entity` passes the shared, already-updated `auth_state["generation"]` to `_reauthenticate` instead of a value snapshotted *before* its own request. When N coroutines hit a 401 from the same expired cookie, the first re-login bumps `auth_state["generation"]`, so every subsequent coroutine reads the new value and its `seen == current` guard passes → each forces another `/security/formLogin`. With a 20-id batch that is up to ~10 concurrent re-logins (per semaphore batch), which can itself trip rate limiting.

The sync path avoids this by snapshotting `self._auth_generation` at the top of each attempt (line 211).

✅ **Fixed in this review:** snapshot `seen_generation = auth_state["generation"]` before the request and pass that to `_reauthenticate`. Verified with a 20-concurrent-401 test → exactly 1 re-login.

vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -223,60 +313,127 @@ def _search_entities(
...
+                            logger.info(f"Fetched {done}/{n_pages} pages of `{endpoint}`")
+                except Exception:
+                    # Completed pages are already checkpointed; a re-run resumes
+                    executor.shutdown(cancel_futures=True)
+                    raise
+
+        results = []
+        for page_offset in sorted(pages):
+            page_results = pages[page_offset].get("results", [])
+            if not page_results:
```
ztawilvdl: **[correctness, high] `break` on an empty interior page silently truncates concurrent results.** Pages are fetched concurrently and assembled in `sorted(pages)` order, but the loop `break`s at the first page whose `results` is empty. Unlike the old strictly-sequential `_paginate` (where an empty page genuinely meant end-of-data), here an interior page can come back empty — a transient HTTP-200-with-`[]`, or rows shifting between concurrent fetches — while higher offsets returned valid rows. The `break` then discards every later page with no error; the caller believes it got everything.

Example: `total_count=250` → offsets 0/100/200; if offset 100 returns `[]`, offset 200's 50 rows are dropped and `count` is reported as 100.

✅ **Fixed in this review:** removed the `break` — empty pages simply contribute nothing. Verified: interior-empty-page scenario now keeps all 150 rows.

vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -87,21 +178,77 @@ def logout(self) -> None:
...
+        jitter. A 401/403 triggers a single re-authentication (the session
+        cookie expires on long runs) before retrying.
+        """
+        url = os.path.join(self.base_url, endpoint)
+        reauthed = False
+        response = None
+        for attempt in range(max_retries):
+            auth_generation = self._auth_generation
+            try:
+                response = self.session.request(
```
ztawilvdl, line 215: **[correctness, high] A single shared `requests.Session` is used from up to 8 `ThreadPoolExecutor` workers at once.** Each page worker calls `self._post` → `_request_with_retries` → `self.session.request(...)` on the same `Session` (created once in `__init__`). `requests.Session` is not documented as thread-safe: its cookie jar is read by every worker to sign requests while a 401-handling worker's `_reauthenticate` → `_authenticate` rewrites that same jar (under `_auth_lock`, which the readers don't hold). Under load this risks intermittent connection-pool churn ("Connection pool is full") and, during a re-auth, half-updated cookies → a cascade of fresh 401s.

⚠️ **Not auto-fixed** — the correct fix (a thread-local `Session`/connection pool per worker, or serializing requests) is a design decision that reaches beyond this diff. Flagging for a maintainer call. At minimum, worth a small live run confirming 8 concurrent page fetches behave before a large crawl.

vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -87,21 +178,77 @@ def logout(self) -> None:
...
+            if response.status_code in RETRYABLE_STATUS_CODES and attempt < max_retries - 1:
+                wait = 2 ** attempt + random.random()
+                logger.warning(f"HTTP {response.status_code} from {endpoint}, retrying in {wait:.1f}s")
+                time.sleep(wait)
+                continue
+
+            response.raise_for_status()
+            return response.json()
+
+        # Only reachable if the final attempt was spent on a re-auth retry
```
ztawilvdl: **[correctness, medium] A 401/403 on the *final* retry attempt re-auths and then raises without ever retrying.** The `for attempt in range(max_retries)` loop treats a re-auth `continue` as consuming an attempt. If attempts 0–2 are spent on retryable 5xx/connection errors and attempt 3 (the last) returns 401, the code re-authenticates, `continue`s, the loop exits, and control falls to the trailing `response.raise_for_status()` — raising the stale 401 even though the freshly-refreshed session would have succeeded on one more try. (The `# Only reachable if the final attempt was spent on a re-auth retry` comment acknowledges this dead-end.)

✅ **Fixed in this review:** converted to a `while` loop where a re-auth does **not** increment `attempt` (bounded by the one-shot `reauthed` latch, so no infinite loop), guaranteeing a real retry after the refresh.

vdl_tools/scrape_enrich/netzero_insights/search_netzero_api.py
```diff
@@ -142,14 +145,29 @@ def search_companies(
     limit: int = 100,
     use_sandbox: bool = False,
     netzero_api: NetZeroAPI = None,
+    sorting: Sorting = None,
+    checkpoint_dir: str = DEFAULT_SEARCH_CHECKPOINT_DIR,
```
ztawilvdl, line 149: **[correctness / data-freshness, medium] `search_companies` now defaults to a 7-day on-disk/S3 result cache, so repeat queries silently return stale data.** With `checkpoint_dir=DEFAULT_SEARCH_CHECKPOINT_DIR` on by default, re-running the same query within `checkpoint_max_age_days` (7) resumes from `meta.json` + cached pages instead of hitting the API — for any `limit > page_size` or `limit=None` call (e.g. `test_login.py:12` uses `limit=None`). Companies added upstream in the last 7 days are missing, with no error and no log at the call site. This is a behavioral shift from the old always-live search and is easy to trip over given VDL's "data lineage is non-negotiable" stance.

⚠️ **Left as-is** (appears intended per the PR description) — but flagging so it's a conscious choice. Consider documenting on the public wrapper that fresh results require `checkpoint_dir=None` or `checkpoint_max_age_days=0`, and note the split behavior at the `limit <= page_size` boundary (that path skips checkpointing entirely, so the default `limit=100` case masks the staleness).

vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -223,60 +313,127 @@ def _search_entities(
...
+            results = deduped
 
-            for page in paginated_results:
-                if total_count is None:
-                    total_count = page.get("count", 0)
-                page_results = page.get("results", [])
-                results.extend(page_results)
-                count_in_result += len(page_results)
+        if limit is not None:
+            results = results[:limit]
```
ztawilvdl, line 442: **[correctness, medium-low] Cross-page dedup can return fewer than `limit` rows even when more distinct results exist.** `n_pages` is computed from the raw `limit` (`ceil(limit/page_size)`), then `unique_key` dedup runs, then `results[:limit]`. If duplicates across page boundaries exceed the last page's slack, the final list is short. E.g. `search_startups(limit=200, page_size=100)` fetches 2 pages (200 raw rows); if dedup drops 30, you get 170 back even though a 3rd page of distinct rows was available. No error — the caller silently under-receives.

Not fixed (proper fix is to keep fetching pages until `limit` *distinct* rows are collected, which changes the fetch shape). Flagging as a known limitation.

vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -223,60 +313,127 @@ def _search_entities(
...
+                    "underlying data changes between runs, resumed pages may "
+                    "contain duplicates or miss rows."
+                )
+            checkpoint = SearchCheckpoint(checkpoint_dir, endpoint, payload, page_size)
+            meta = checkpoint.load_meta(timedelta(days=checkpoint_max_age_days))
+            if meta:
+                logger.info(f"Resuming search from checkpoint {checkpoint.base_uri}")
+
+        def fetch_page(page_offset: int, allow_cached: bool) -> Dict:
+            if allow_cached and checkpoint and checkpoint.has_page(page_offset):
```
ztawilvdl: **[efficiency] Redundant existence probe before every checkpoint read (~2 extra S3 HEADs per page on resume).** `fetch_page` does `checkpoint.has_page()` (→ `target_exists` = `bucket_exists()` + `key_exists()`, two S3 HEADs) and then `read_page()` (which does its own ETag HEAD). `load_meta` has the same `target_exists` + `read_json` pattern. On a 500-page S3 resume that's ~1000 wasted round trips, inside the thread pool.

✅ **Fixed in this review:** read straight through and catch `FileNotFoundError` to fall back to the API — one code path instead of probe-then-read. (`SearchCheckpoint.has_page` is now unused; left in place as a small public helper.)

vdl_tools/scrape_enrich/netzero_insights/netzero_api.py
```diff
@@ -223,60 +313,127 @@ def _search_entities(
...
+                checkpoint.write_page(page_offset, page)
+            return page
+
+        # Only trust existing page files when resuming a fresh checkpoint —
+        # otherwise stale pages from an expired run would be read back
+        resuming = meta is not None
+
+        # First page establishes total_count (from checkpoint meta when resuming)
+        first_page = fetch_page(offset, allow_cached=resuming)
+        total_count = meta["total_count"] if resuming else first_page["count"]
```
ztawilvdl, line 385: **[correctness, low/speculative] `total_count = data.get("count", 0)` silently under-fetches if the API ever omits/renames `count`.** On the first page, a missing `count` → `total_count=0` → `available=0` → `n_pages=1`, so only the first 100 rows are returned for a 500-row result set, and the method reports `total_count=0` with no exception. Resilient degradation would be better than silent truncation here (e.g. fall back to paging until an empty page when `count` is absent).

Not fixed (depends on an API-schema assumption I can't verify without live calls) — flagging for awareness.

PR #156, opened 2026-07-16
Separate cache model_name from the API model; add reasoning to classi…

larareichmann:
…fy_entities

Reasoning effort changes the response but is not part of the cache key (prompt_id, given_id, model_name), so two runs differing only in effort collide. Add an optional cache_model_name that tags the cache row while the API is still called with the real model, letting a run be namespaced as e.g. "gpt-5.4-mini#reasoning=medium".

- prompt_response_cache_sql: cache_model_name param, applied at the four cache-key sites only; self.model still drives the API call and token sizing.
- prompt_response_cache_instructor / taxonomy_mapping_cache: forward it through.
- hierarchical_taxonomy_mapping: reasoning + cache_model_name on classify_entities, both guarded on is_reasoning_model. Warn when a reasoning model runs with no explicit effort, since the model's silent default then applies and is not recorded in the key (gpt-5-mini defaults to medium, gpt-5.4-mini to none). Warn rather than assume a default table: the defaults are OpenAI's to change, and a key asserting the wrong effort is worse than none.
- openai_constants: add the gpt-5-mini MODEL_DATA entry get_context_window needs.

All new params are optional and default to existing behavior.

### DIFF
--- a/vdl_tools/shared_tools/openai/openai_constants.py
+++ b/vdl_tools/shared_tools/openai/openai_constants.py
@@ -1,4 +1,11 @@
 MODEL_DATA = {
+    "gpt-5-mini": {
+        "model_name": "gpt-5-mini",
+        "max_context_window": 400_000,
+        "max_output_tokens": 128_000,
+        "input_cost_per_token": 0.25 / 1_000_000,
+        "output_cost_per_token": 2.0 / 1_000_000,
+    },
     "gpt-5.5": {
         "model_name": "gpt-5.5",
         "max_context_window": 1_000_000,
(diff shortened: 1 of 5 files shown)


ztawilvdl:
Superseded by #157, which implements the systematic version of this idea from Step 4 of the cache refactor plan: instead of a manually tagged `cache_model_name`, the cache now computes a `request_hash` from an allowlist of output-affecting kwargs (`reasoning`, `tools`, `temperature`, …) and includes it in the PK — so effort-only differences get separate rows automatically, with no per-call-site tagging to forget.

Two pieces of this PR live on in #157 (thank you, @larareichmann 🙌): the `gpt-5-mini` `MODEL_DATA` entry, and the `reasoning` passthrough on `classify_entities` — generalized to `**api_kwargs` so any future output-affecting kwarg flows through. The warning about reasoning models running with an unpinned default effort also survives in condensed form.

PR #157, opened 2026-07-16
Cache: request_hash kwargs-aware cache keys (Step 4 of cache refactor)

ztawilvdl:
## What

We now treat a model as its name **plus** the hyperparameters sent with the call. The cache computes a `request_hash` per call from an explicit allowlist of output-affecting kwargs (`reasoning`, `tools`, `tool_choice`, `temperature`, `top_p`, `max_output_tokens`, `seed`, `service_tier`) and adds it to the `prompt_response` PK, so e.g. `reasoning={"effort": "low"}` vs `{"effort": "high"}` occupy separate rows instead of silently colliding.

Design points:
- **`request_hash` pairs with `model_name` under `filter_by_model`** — the two form one "model identity". `filter_by_model=True` keys reads on both name and kwargs (set it whenever you A/B models *or* hyperparameters); `False` (default) shares rows across model identities entirely, exactly preserving legacy read behavior. Writes always stamp both (they're in the PK).
- **`text_format` is excluded** — `response_full` is stored raw and parsed at retrieval, so changing the Pydantic response model reuses the row.
- **Legacy compatible**: calls with no allowlisted kwargs hash to `''`, which is also the migration backfill. Combined with the pairing above, default-flag callers take **no** cache bust from this change.
- **`request_kwargs` JSONB column** stores the normalized kwargs alongside the hash for auditability/lineage.
- `classify_entities` / `recover_unmatched` (and `add_hierarchical_taxonomy_mapping`) take an explicit `llm_api_kwargs` dict (e.g. `{"reasoning": {"effort": "low"}}`) forwarded verbatim to the API and the cache key — explicit rather than `**kwargs` so a typo'd function kwarg fails loudly at the call instead of surfacing as cache error rows, with a warning when a reasoning model runs with no explicit effort (the model's silent default isn't recorded in the key).

Supersedes #156 — thanks @larareichmann for surfacing the collision; the `gpt-5-mini` `MODEL_DATA` entry and the reasoning passthrough from that PR are folded in here (generalized to the explicit `llm_api_kwargs` dict). The manual `cache_model_name` mechanism is dropped in favor of the automatic hash, and the pairing with `filter_by_model` matches her original gating.

## Deployment

Migration `36a8b8005ff8` adds the two columns and widens the PK (4.5M rows — brief lock during PK rebuild). Old code + new schema is safe (`server_default=''`); new code + old schema is **not** — merge and run `alembic upgrade head` together.

Known one-time cache busts (accepted, discussed with Zein): limited to `filter_by_model=True` callers that pass allowlisted kwargs (e.g. taxonomy drivers defaulting `filter_by_model=True` with `temperature=0`).

## Verification (scratch DB + real API)

- Migration: `upgrade` → `downgrade` (dedupes) → `upgrade`, legacy row backfills to `''`.
- `filter_by_model=True`: `reasoning` low vs high → two rows with distinct hashes and `request_kwargs` populated; identical kwargs → cache hit with 0 API calls; different kwargs → all miss (API attempted); planted pre-hash `''` row hits a no-kwargs call.
- `filter_by_model=False`: reads ignore kwargs — single row, cross-effort cache hit (legacy sharing preserved).
- `text_format` A → B → single row, second call a hit.
- Bulk + error paths: rows carry hash+kwargs; error rows upsert on the 4-col key and increment `num_errors`.
- `classify_entities` e2e with `filter_by_model=True` and `reasoning` via `llm_api_kwargs`: pass 1 classifies, pass 2 fully cached.

### DIFF
--- a/vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
+++ b/vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
@@ -1,0 +1,79 @@
+"""add request_hash and request_kwargs to prompt_response
+
+Adds kwargs-aware cache keys (CACHE_REFACTOR_PLAN.md Step 4):
+
+- `request_hash`: hash of the allowlisted output-affecting API kwargs
+  (reasoning, tools, temperature, ...). Existing rows backfill to '' via the
+  server default, which is also what calls with no allowlisted kwargs hash
+  to — so legacy rows keep matching those calls.
+- `request_kwargs`: the normalized kwargs the hash was computed from, kept
+  for auditability.
+- PK widens from (prompt_id, given_id, model_name) to include request_hash,
+  so calls differing only in output-affecting kwargs occupy separate rows.
+
+Revision ID: 36a8b8005ff8
+Revises: 8fa633af5bbd
+Create Date: 2026-07-16 12:20:49.752595
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = '36a8b8005ff8'
+down_revision: Union[str, None] = '8fa633af5bbd'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # server_default='' backfills all existing rows in one pass, and keeps
+    # inserts from deployments that predate the ORM change valid.
+    op.add_column('prompt_response', sa.Column('request_hash', sa.String(), server_default='', nullable=False))
+    op.add_column('prompt_response', sa.Column('request_kwargs', postgresql.JSONB(astext_type=sa.Text()), nullable=True))
+    op.create_index(op.f('ix_prompt_response_request_hash'), 'prompt_response', ['request_hash'], unique=False)
+
+    # Widen the PK to include request_hash (same pattern as 223e827c20cd,
+    # which widened it to include model_name).
+    op.drop_constraint('prompt_response_pkey', 'prompt_response', type_='primary')
+    op.create_primary_key(
+        'prompt_response_pkey',
+        'prompt_response',
+        ['prompt_id', 'given_id', 'model_name', 'request_hash'],
+    )
+
+
+def downgrade() -> None:
+    op.drop_constraint('prompt_response_pkey', 'prompt_response', type_='primary')
+
+    # Rows that differ only in request_hash collide under the narrower PK —
+    # keep the most recently updated one per (prompt_id, given_id, model_name).
+    op.execute("""
+        DELETE FROM prompt_response pr
+        USING (
+            SELECT prompt_id, given_id, model_name, request_hash,
+                   ROW_NUMBER() OVER (
+                       PARTITION BY prompt_id, given_id, model_name
+                       ORDER BY date_updated DESC NULLS LAST,
+                                date_added DESC NULLS LAST
+                   ) AS rn
+            FROM prompt_response
+        ) ranked
+        WHERE pr.prompt_id = ranked.prompt_id
+          AND pr.given_id = ranked.given_id
+          AND pr.model_name = ranked.model_name
+          AND pr.request_hash = ranked.request_hash
+          AND ranked.rn > 1
+    """)
+
+    op.drop_index(op.f('ix_prompt_response_request_hash'), table_name='prompt_response')
+    op.drop_column('prompt_response', 'request_kwargs')
+    op.drop_column('prompt_response', 'request_hash')
+    op.create_primary_key(
+        'prompt_response_pkey',
+        'prompt_response',
+        ['prompt_id', 'given_id', 'model_name'],
+    )
--- a/vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
+++ b/vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
@@ -30,11 +30,18 @@ depends_on: Union[str, Sequence[str], None] = None
 
 
 def upgrade() -> None:
-    # server_default='' backfills all existing rows in one pass, and keeps
-    # inserts from deployments that predate the ORM change valid.
+    # server_default='' backfills all existing rows in one pass and keeps
+    # plain INSERTs from pre-PR code valid. It does NOT make old code safe
+    # against the new schema: pre-PR bulk upserts use ON CONFLICT
+    # (prompt_id, given_id, model_name), which stops matching any unique
+    # index once the PK widens below (Postgres 42P10). Deploy code and this
+    # migration together, with no old writers still running.
+    #
+    # No standalone index on request_hash: no query filters by it alone
+    # (reads always pair it with indexed prompt_id/text_id/given_id), and
+    # the column is '' for virtually all rows.
     op.add_column('prompt_response', sa.Column('request_hash', sa.String(), server_default='', nullable=False))
     op.add_column('prompt_response', sa.Column('request_kwargs', postgresql.JSONB(astext_type=sa.Text()), nullable=True))
-    op.create_index(op.f('ix_prompt_response_request_hash'), 'prompt_response', ['request_hash'], unique=False)
 
     # Widen the PK to include request_hash (same pattern as 223e827c20cd,
     # which widened it to include model_name).
@@ -69,7 +76,6 @@ def downgrade() -> None:
           AND ranked.rn > 1
     """)
 
-    op.drop_index(op.f('ix_prompt_response_request_hash'), table_name='prompt_response')
     op.drop_column('prompt_response', 'request_kwargs')
     op.drop_column('prompt_response', 'request_hash')
     op.create_primary_key(
--- a/vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
+++ b/vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
@@ -101,6 +101,20 @@ def _request_hash(norm_kwargs: dict) -> str:
     return create_deterministic_md5(json.dumps(norm_kwargs, sort_keys=True))
 
 
+def _row_preference(row) -> tuple:
+    """Sort key for picking one row when several share (given_id, text_id).
+
+    Reads at filter_by_model=False can match one row per model identity
+    (name + request_hash) under the widened PK. A success row always beats
+    an error row — a failed run under one identity must not shadow another
+    identity's cached success — and newer beats older.
+    """
+    return (
+        not row.num_errors,
+        row.date_updated or row.date_added or dt.datetime.min,
+    )
+
+
 def _resolve_cache_flags(
     read_from_cache: bool,
     write_to_cache: bool,
@@ -459,6 +473,17 @@ class PromptResponseCacheSQL():
             self.session
             .query(PromptResponse)
             .filter(*filters)
+            # At filter_by_model=False several rows can match (one per model
+            # identity under the widened PK); without an ORDER BY, .first()
+            # is arbitrary. Deterministic winner: success rows before error
+            # rows, then most recently touched (date_updated is NULL until a
+            # row is refreshed, so fall back to date_added).
+            .order_by(
+                func.coalesce(PromptResponse.num_errors, 0).asc(),
+                func.coalesce(
+                    PromptResponse.date_updated, PromptResponse.date_added
+                ).desc().nullslast(),
+            )
         )
         return prompt_response_obj.first()
 
@@ -512,8 +537,19 @@ class PromptResponseCacheSQL():
 
         logger.info("Found %s total rows", len(found_rows))
 
-        found_rows_to_errors = {(x.given_id, x.text_id): x.num_errors for x in found_rows}
-        found_rows = [x for x in found_rows if not found_rows_to_errors.get((x.given_id, x.text_id))]
+        # At filter_by_model=False several rows can match one (given_id,
+        # text_id) — one per model identity under the widened PK. Collapse
+        # via _row_preference (success beats error, newer beats older) so an
+        # error row never shadows a cached success out of the results.
+        best_rows: dict[tuple[str, str], PromptResponse] = {}
+        for row in found_rows:
+            key = (row.given_id, row.text_id)
+            prev = best_rows.get(key)
+            if prev is None or _row_preference(row) > _row_preference(prev):
+                best_rows[key] = row
+
+        found_rows_to_errors = {key: row.num_errors for key, row in best_rows.items()}
+        found_rows = [row for row in best_rows.values() if not row.num_errors]
 
         found_rows_keys = found_rows_to_errors.keys()
         unfound_ids_or_errors = {
--- a/vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
+++ b/vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
@@ -23,11 +23,14 @@ from sqlalchemy import select, func, tuple_
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from more_itertools import chunked
 
-from vdl_tools.shared_tools.database_cache.database_models.prompt import Prompt, PromptResponse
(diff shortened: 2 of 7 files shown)


vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
```diff
@@ -0,0 +1,79 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '36a8b8005ff8'
+down_revision: Union[str, None] = '8fa633af5bbd'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # server_default='' backfills all existing rows in one pass, and keeps
+    # inserts from deployments that predate the ORM change valid.
```
ztawilvdl: **[code-review] The “old code + new schema is safe” claim doesn’t hold for bulk upserts.**

Pre-PR code’s `_upsert_success_rows`/`_upsert_error_rows` use `on_conflict_do_update(index_elements=['prompt_id','given_id','model_name'])`. Postgres ON CONFLICT inference requires an exact unique-index match, so once this migration widens the PK to 4 columns, every bulk cache write from a still-running pre-PR deployment fails with 42P10 “there is no unique or exclusion constraint matching the ON CONFLICT specification”. `server_default=''` only rescues plain INSERTs and `session.merge` single-row paths.

Suggest correcting this comment (and the plan doc’s “Old code + new schema is safe” line) — the coordination constraint is bidirectional: old bulk writers must be stopped before `alembic upgrade head`.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
```diff
@@ -409,6 +502,7 @@ def get_prompt_response_obj_bulk(
 
         if self.filter_by_model:
             filters.append(PromptResponse.model_name == self.model)
+            filters.append(PromptResponse.request_hash == request_hash)
```
ztawilvdl: **[code-review] Under default `filter_by_model=False`, an error row can now shadow a cached success in the bulk read, silently dropping the entity.**

The widened PK lets a success row (`request_hash=''`) and an error row (non-empty hash) coexist for the same model + `(given_id, text_id)` — impossible pre-PR, when the second write upserted onto the same row. The default-path bulk query applies neither filter, returns both rows, and `found_rows_to_errors = {(x.given_id, x.text_id): x.num_errors ...}` collapses them in arbitrary order. If the error row iterates last, the success row is excluded from `found_rows`, and with default `max_errors=1` the retry gate also skips the item — the given_id vanishes from the returned dict with no error.

Consider preferring success rows (lowest `num_errors`) when collapsing duplicates instead of arbitrary last-write-wins.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
```diff
@@ -366,6 +453,7 @@ def get_prompt_response_obj(self, given_id: str, text: str):
         ]
         if self.filter_by_model:
             filters.append(PromptResponse.model_name == self.model)
+            filters.append(PromptResponse.request_hash == request_hash)
```
ztawilvdl: **[code-review] Reads at default `filter_by_model=False` ignore `request_hash` while writes key on it — `.first()` with no ORDER BY then serves an arbitrary, possibly stale row.**

Pre-PR, a same-model refresh (`read_from_cache=False`) reliably overwrote the single row. Now a refresh with different allowlisted kwargs INSERTs a second row under the widened PK, the stale row survives, and every subsequent default-config read nondeterministically returns either one. The docstring documents cross-identity *sharing*, but not arbitrary/stale selection among coexisting rows.

Consider adding a deterministic ORDER BY (e.g. `date_updated DESC`) to both the single and bulk lookups.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -731,11 +749,18 @@ def classify_entities(
     empty_leaf = {k: None for k in leaf_keys}
     response_model: type[BaseModel] = match_schema or MatchesResponse
 
+    # Copy — the caller's dict must not grow a temperature key.
+    api_kwargs: dict[str, Any] = dict(llm_api_kwargs or {})
```
ztawilvdl: **[code-review] The two temperature channels interact badly here (and at line 1601 in `recover_unmatched`).**

1. The `temperature` param defaults to `0`, so the guarded assignment below silently clobbers a caller-supplied `llm_api_kwargs={'temperature': 0.7}` on non-reasoning models — contradicting the docstring’s “forwarded verbatim”, and the cache row’s `request_hash`/`request_kwargs` record `temperature=0`, a request the caller never made.
2. The reasoning-model suppression only guards the param channel: a `temperature` inside `llm_api_kwargs` reaches gpt-5* models unstripped, so every call 400s and the run completes writing only error rows — despite the docstring promising temperature is “automatically suppressed for reasoning models”.

Suggest raising (or warning) when `temperature` appears in both channels, and applying the reasoning-model strip to the merged dict — ideally in a shared helper, since this block is duplicated in `recover_unmatched` and the two copies already diverge (the no-reasoning warning exists only here).

vdl_tools/shared_tools/taxonomy_mapping/few_shot_cache.py
```diff
@@ -476,7 +476,7 @@ def get_completion(
 
         return response
 
-    def _build_success_row(self, given_id, text, response):
+    def _build_success_row(self, given_id, text, response, request_hash="", request_kwargs=None):
```
ztawilvdl: **[code-review] `FewShotCache.get_completion` never forwards `**kwargs` to `CLIENT.responses.parse`, but the base class now hashes those same kwargs into `request_hash` and persists them as `request_kwargs`.**

The base `get_completion` forwards kwargs to the API; this subclass’s override builds `response_kwargs` from only model/input/text_format/instructions and drops the rest (its own docstring claims they’re “Passed to the Responses API”). So `reasoning={'effort': 'low'}` vs `'high'` through this cache pays for two byte-identical API calls stored as two rows whose `request_kwargs` record efforts that were never sent — the audit column added by this PR would lie for every row written through this class.

The kwargs drop predates this PR, but the PR makes it actively harmful; merge the kwargs into `response_kwargs` before the `CLIENT.responses.parse` call.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
```diff
@@ -619,6 +728,7 @@ def store_error(
             .filter(
                 PromptResponse.prompt_id == self.prompt.id,
                 PromptResponse.text_id == text_id,
+                PromptResponse.request_hash == request_hash,
```
ztawilvdl: **[code-review] `store_error`’s previous-row lookup now filters `request_hash` but still not `model_name` — the “model identity = name + hash” pairing is only half applied on this write path.**

Under the 4-column PK, rows for the same prompt/text/hash but different models legitimately coexist; the unordered `.first()` can grab another model’s row, overwrite its `response_full` with this model’s error payload and bump its `num_errors` — while the failing model’s own count stays 0, so its retries are never capped by `max_errors`.

Add `PromptResponse.model_name == self.model` here (and the same in `few_shot_cache.store_error`, which filters `given_id` + `request_hash` only).

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
```diff
@@ -55,6 +57,50 @@
...
+
+    Values are round-tripped through JSON with sorted keys so structural
+    equality maps to the same normalized form (e.g. dict key order or tool
+    list contents don't produce spurious cache misses).
+    """
+    norm = {}
+    for key in sorted(kwargs):
+        if key not in KWARG_KEYS_THAT_AFFECT_OUTPUT:
+            continue
+        norm[key] = json.loads(json.dumps(kwargs[key], sort_keys=True, default=str))
```
ztawilvdl: **[code-review] `default=str` makes `request_hash` depend on Python reprs.**

A non-JSON-serializable value in an allowlisted kwarg (the allowlist invites `tools`/`tool_choice`) falls through to `str()`; address-bearing reprs like `<function f at 0x104ab3d80>` differ every process, so the hash never matches prior runs — under `filter_by_model=True` that’s a 100% cache miss per run (full API re-spend) plus a new PK row per item per run, and `request_kwargs` stores junk reprs instead of the real payload. Latent today (in-repo callers pass JSON-native values only), but cheap to guard: drop `default=str` and let a non-serializable kwarg raise loudly at the call site — consistent with this PR’s own fail-loudly rationale for `llm_api_kwargs`.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
```diff
@@ -1395,6 +1428,7 @@ def add_hierarchical_taxonomy_mapping(
             match_schema=match_schema,
             temperature=temperature,
             filter_by_model=filter_by_model,
+            llm_api_kwargs=llm_api_kwargs,
```
ztawilvdl: **[code-review] `llm_api_kwargs` is forwarded to the injected `classify_fn` unconditionally — every external `classify_fn` written to the pre-PR contract now raises `TypeError`, even with the feature unused.**

With `llm_api_kwargs=None` and a custom `classify_fn` that predates this PR, the call dies with `TypeError: ... got an unexpected keyword argument 'llm_api_kwargs'`. The docstring documents the new contract, but forwarding only when `llm_api_kwargs is not None` keeps existing injected fns (e.g. drawdown-style drivers) working at no cost.

vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
```diff
@@ -0,0 +1,79 @@
...
+    op.add_column('prompt_response', sa.Column('request_hash', sa.String(), server_default='', nullable=False))
+    op.add_column('prompt_response', sa.Column('request_kwargs', postgresql.JSONB(astext_type=sa.Text()), nullable=True))
+    op.create_index(op.f('ix_prompt_response_request_hash'), 'prompt_response', ['request_hash'], unique=False)
```
ztawilvdl: **[code-review] This standalone index can’t be used by any existing query.**

Every read filtering `request_hash` also filters already-indexed `prompt_id`/`text_id`/`given_id` (and the new PK btree covers it as trailing column), and the column is `''` for virtually all rows — near-zero cardinality, so the planner will never choose it. It’s pure btree maintenance + WAL on every write to a high-write cache table.

Suggest dropping `op.create_index`/`op.drop_index` here and `index=True` on the ORM column. (`ix_prompt_response_model_name` from 223e827c20cd has the same problem — precedent, but not a reason to repeat it.)

vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Widen the PK to include request_hash (same pattern as 223e827c20cd,
+    # which widened it to include model_name).
+    op.drop_constraint('prompt_response_pkey', 'prompt_response', type_='primary')
```
ztawilvdl, line 48: **[code-review] Constraint-name mismatch with the ORM (pre-existing, low severity).**

All alembic-lineage DBs are fine — 223e827c20cd explicitly created the PK as `prompt_response_pkey`. But the ORM names the same PK `unique_prompt_given_id_model_name` (`database_models/prompt.py`), so a DB bootstrapped via `Base.metadata.create_all` (`database_utils.recreate_db`) and stamped to head fails here with “constraint prompt_response_pkey does not exist” (the migration transaction rolls back cleanly). Consider aligning the ORM constraint name with the migration lineage.

vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
ztawilvdl: Fixed in e372b41 — the migration comment and CACHE_REFACTOR_PLAN.md now state the deploy constraint as bidirectional: old bulk writers must stop before `alembic upgrade head`.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
ztawilvdl: Fixed in c20f2de — the bulk collapse now picks a deterministic best row per (given_id, text_id) via `_row_preference` (success beats error, newer beats older), so an error row can no longer shadow a cached success. Verified e2e against Postgres.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
ztawilvdl: Fixed in c20f2de — the single read now has a deterministic ORDER BY: success rows first (`coalesce(num_errors, 0) ASC`), then most recently touched (`coalesce(date_updated, date_added) DESC NULLS LAST`). Verified e2e against Postgres.

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
ztawilvdl: Fixed in e372b41 — both call sites now share `_build_llm_api_kwargs`: a `temperature` inside `llm_api_kwargs` wins over the parameter default, and is stripped (with a warning) for reasoning models. The no-reasoning-effort warning now fires in `recover_unmatched` too.

vdl_tools/shared_tools/taxonomy_mapping/few_shot_cache.py
ztawilvdl: Fixed in e372b41 — `get_completion` now merges `**kwargs` into the `responses.parse` call, so `request_hash`/`request_kwargs` describe the request actually made. Verified with a fake client that `reasoning`/`temperature`/`seed` arrive and the reasoning branch still sets instructions/input.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
ztawilvdl: Fixed in e372b41 — both `store_error` lookups now filter `model_name` alongside `request_hash`. Verified e2e: model B failing twice increments only its own row; model A's cached success is untouched.

vdl_tools/shared_tools/openai/prompt_response_cache_sql.py
ztawilvdl: Fixed in e372b41 (dropped `default=str`; non-serializable allowlisted kwargs now raise TypeError at the call site) and hardened in be07f98, which moved all key derivation onto `PromptResponse` (`normalize_request_kwargs`/`create_request_hash`, mirroring `text_id`).

vdl_tools/shared_tools/taxonomy_mapping/hierarchical_taxonomy_mapping.py
ztawilvdl: Fixed in e372b41 — `llm_api_kwargs` is forwarded to an injected `classify_fn` only when supplied, so pre-existing custom `classify_fn` implementations keep working until a caller uses the feature.

vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
ztawilvdl: Fixed in e372b41 — `op.create_index`/`op.drop_index` removed from the migration and `index=True` dropped from the ORM column. Note: if 36a8b8005ff8 already ran on a live DB, drop `ix_prompt_response_request_hash` there manually.

vdl_tools/shared_tools/database_cache/data_migrations/versions/36a8b8005ff8_add_request_hash_and_request_kwargs_to_.py
ztawilvdl: Fixed in e372b41 — the ORM `PrimaryKeyConstraint` is renamed to `prompt_response_pkey` to match the alembic lineage (223e827c20cd). Verified e2e that `create_all` now produces that name.

